PR #12, opened 2023-09-27
fix: dynamic version not working with "dev mode" installs.

fbdtemme:
# Description

Fix an issue where editable installations with pip and poetry would still return the dynamic versioning placeholder `0.0.0` instead of the real version.

## Type of change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,9 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-* Fix an issue where `--keep-workdirs` option for pytest was not available when running pytest without
-  restricting the testdir to `tests/integration`.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -24,9 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-* Fix an issue where `--keep-workdirs` option for pytest was not available when running pytest without
```
ambarrio, line 27: Why are we taking this out?

ambarrio (approved):
I approve, but can you check why we take out one of the fixed lines in the changelog?

CHANGELOG.md
fbdtemme: This was already in 0.13.1 and was just copied over.

PR #25, opened 2023-10-16
Fix sample column missing from concatenated edgelist

johandahlberg:
## Description

The sample name was missing from the concatenated edge lists. This bug was introduced when I switched to using polars for concatenating the edge lists. This fixes that, and makes sure that the tests catches this problem.

Fixes: EXE-1174

## Type of change

## How Has This Been Tested?

Tested with the unit test suite, including new test code to verify that this bug has been fixed.


ambarrio (approved):
LGTM

PR #28, opened 2023-10-20
Fixes to pixeldataset aggregation and parquet file indexeing

johandahlberg:
## Description

Fixes the following issues:
 - aggregation of more than two pixeldatasets was broken.
 - indexes were written to the parquet files, which caused downstream issues.

Fixes: EXE-1184

## Type of change

## How Has This Been Tested?

With unit tests, as well as by manually verifying that it works in the scenario where the bug was originally reported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-* Fixed broken pixeldataset aggregation for more than two sample
-* Fixed bug in graph generation caused by accidentally written index in parquet file.
+* Fixed broken pixeldataset aggregation for more than two samples.
+* Fixed a bug in graph generation caused by accidentally writing the index to the parquet file.
   For backwards compatiblity, if there is an column named `index`` in the edgelist, this
   will be removed and the user will get a warning indicating that this has happened.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [UNRELEASED] - ...
+
+### Fixed
+
+* Fixed broken pixeldataset aggregation for more than two sample
+* Fixed bug in graph generation caused by accidentally written index in parquet file.
+  For backwards compatiblity, if there is an column named `index`` in the edgelist, this
+  will be removed and the user will get a warning indicating that this has happened.
+
 ## [0.15.1] - 2023-10-18
 
 ### Fixed
--- a/tests/test_pixeldataset.py
+++ b/tests/test_pixeldataset.py
@@ -13,6 +13,7 @@ from unittest import mock
 
 import numpy as np
 import pandas as pd
+import polars as pl
 import pytest
 from anndata import AnnData
 from numpy.testing import assert_array_equal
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [UNRELEASED] - ...
+
+### Fixed
+
+* Fixed broken pixeldataset aggregation for more than two sample
```
fbdtemme, line 12:
```suggestion
* Fixed broken pixeldataset aggregation for more than two samples.
```

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
...
+* Fixed bug in graph generation caused by accidentally written index in parquet file.
```
fbdtemme, line 13:
```suggestion
* Fixed a bug in graph generation caused by accidentally writing the index to the parquet file.
```

fbdtemme (approved):
LGTM

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
...
+  For backwards compatiblity, if there is an column named `index`` in the edgelist, this
```
ambarrio, line 14: `index`` -> `index`

tests/test_pixeldataset.py
```diff
@@ -70,6 +71,22 @@ def test_pixel_file_parquet_format_spec_can_save(setup_basic_pixel_dataset, tmp_
...
+    # read with pandas or polars
+    file_target = tmp_path / "df.parquet"
+    df = pd.DataFrame({"A": [1, 2, 3], "B": [1, 2, 3]}, index=["X", "Y", "Z"])
+    PixelFileParquetFormatSpec().serialize_dataframe(df, path=file_target)
+    assert file_target.is_file()
+    res1 = pd.read_parquet(file_target)
+    assert set(res1.columns) == {"A", "B"}
+    res2 = pl.scan_parquet(file_target)
+    assert set(res2.columns) == {"A", "B"}
+    res3 = pl.read_parquet(file_target)
```
ambarrio, line 86: WHat are we testing here that is different to line 82?

tests/test_pixeldataset.py
```diff
@@ -715,6 +732,107 @@ def test_filter_should_return_proper_typed_edgelist_data(setup_basic_pixel_datas
...
+    file_target_1 = tmp_path / "dataset_1.pxl"
+    dataset_1.save(str(file_target_1))
+    file_target_2 = tmp_path / "dataset_2.pxl"
+    dataset_2.save(str(file_target_2))
+    file_target_3 = tmp_path / "dataset_3.pxl"
+    dataset_3.save(str(file_target_3))
+    file_target_4 = tmp_path / "dataset_4.pxl"
+    dataset_4.save(str(file_target_4))
+
+    result = simple_aggregate(
```
ambarrio, line 775: I don't understand what is what makes it fail above with two datasets in L747 and it works seamlessly here. Can you explain?

ambarrio: There is not enough info in the test to understand what is different.

ambarrio (approved):
The PR is merged, so nothing blocking. However, questions and cosmetic changes remain to be taken.

I would also like to understand the testing logic on this. It doesn't come across in the tests created. But we can take that apart.

CHANGELOG.md
johandahlberg: I'll fix this up when I prep the release.

tests/test_pixeldataset.py
johandahlberg: One reads with polars and one with pandas (`pl` vs. `pd`).

tests/test_pixeldataset.py
johandahlberg: I think it's the opposite. It works with two samples, but fails when there was more than two samples. That is what this checks for. I can add some more info on this in a comment to make it clear.

tests/test_pixeldataset.py
ambarrio: Ah ya!

PR #30, opened 2023-10-23
Update devcontainer to use py3.11

johandahlberg:
## Description

Make the vscode devcontainer use python 3.11, and fix up some other random issues with it.


ambarrio (approved):
LGTM

PR #32, opened 2023-10-25
Feature/exe 1096 graph building in networkx

johandahlberg:
## Description

This includes building the networkx graph using networkx, as well as some related utility functions.

I've also refactored the tests a bit to make sure tests are in the test files corresponding to the file that the src is in. And I've also made it possible to use `pl.LazyFrame` when building the graph from the edge list. This is only really used by the networkx implementation at the moment, but if we wanted to we could change the igraph backend to also use lazy edge lists to save some memory.

Fixes: EXE-1096

## Type of change

## How Has This Been Tested?

Tested with the unit tests, and I've also tried it on some data locally.

## PR checklist:


src/pixelator/graph/backends/implementations.py
```diff
@@ -428,9 +517,104 @@ def __init__(
...
+    def _build_graph_with_node_counts_from_edgelist(
+        df: pl.LazyFrame,
+        create_using: bool,
+    ) -> Union[nx.Graph, nx.MultiGraph]:
+        unique_markers = set(df.unique("marker").collect()["marker"].to_list())
+        initial_marker_dict = {marker: 0 for marker in unique_markers}
+
+        def _add_or_append_to_node(g, node, marker):
+            if g.nodes.get(node):
+                g.nodes.get(node)["markers"][marker] += 1
```
fbdtemme, line 544: Maybe save the extra lookup here since this is quite performance sensitive?
```suggestion
            existing_node = g.nodes.get(node)
            if existing_node:
                existing_node["markers"][marker] += 1
```

johandahlberg: Good call!

src/pixelator/graph/backends/implementations.py
```diff
@@ -428,9 +517,104 @@ def __init__(
...
+
+        g: nx.Graph = nx.empty_graph(0, create_using)
+
+        for row in (
+            df.select(["upia", "upib", "marker"])
+            .collect(streaming=True)
+            .iter_rows(named=False, buffer_size=1000)
+        ):
+            _add_or_append_to_node(g, row[0], row[2])
+            _add_or_append_to_node(g, row[1], row[2])
```
fbdtemme, line 559: I was also thinking that maybe inlining the `_add_or_append_to_node` nested function here could help a bit. Not sure though if it will matter that match to justify the hit in readability, maybe something for a quick benchmark?

fbdtemme (approved):

src/pixelator/graph/backends/implementations.py
johandahlberg: It seems to give a performance boost of roughly 10% in the microbenchmark - what do you think? Is it worth it?

fbdtemme: I think that is definitly worth it, the function is quite small so it remains quite imho.

johandahlberg: Ok. I've pushed it now. 😄

tests/graph/test_graph.py
```diff
@@ -61,124 +54,166 @@ def create_simple_edge_list_from_graph(
...
+    """Build an A-node projected graph."""
+    graph = Graph.from_edgelist(
+        edgelist=full_graph_edgelist,
+        add_marker_counts=True,
+        simplify=True,
+        use_full_bipartite=False,
     )
-    assert result == [{"CTCGTACCTGGGACTGATACT", "TGTAAGTCAGTTGCAGGTTGG"}]
+    assert graph.vcount() == 50
+    assert graph.ecount() == ((50 * 50) / 2) - (50 / 2)
```
ambarrio, line 168: Every time I look at this test I wonder how do we enforce this. And I have to look at the definition of `full_graph_edgelist` that is a fully connected bipartite graph. I wonder if we could put that into something. A comment before this assert or something like that.

tests/graph/test_graph.py
```diff
@@ -61,124 +54,166 @@ def create_simple_edge_list_from_graph(
...
-    """Test discovery of crossing edges from graph with no communities."""
-    result = community_detection_crossing_edges(
-        graph=graph_without_communities,
-        leiden_iterations=2,
+@pytest.mark.parametrize("enable_backend", ["igraph", "networkx"], indirect=True)
+def test_build_graph_a_node_projected_benchmark(
+    benchmark, enable_backend, full_graph_edgelist: pd.DataFrame
+):
+    """Build an A-node projected graph."""
+    graph = benchmark(
```
ambarrio, line 179: Curious to know how is benchmark used here.

tests/graph/test_graph.py
```diff
@@ -61,124 +54,166 @@ def create_simple_edge_list_from_graph(
...
+            simplify=False,
+            use_full_bipartite=False,
+        )
+        assert graph.vcount() == 50
+        assert graph.ecount() == ((50 * 50) / 2) - (50 / 2)
+        assert "markers" in graph.vs.attributes()
+        assert sorted(list(graph.vs[0]["markers"].keys())) == ["A", "B"]
+        assert graph.vs.attributes() == {"name", "markers", "type", "pixel_type"}
+
+    # We want to warn when these to there is an a-node projection
```
ambarrio, line 213: ????

src/pixelator/graph/backends/implementations.py
```diff
@@ -6,12 +6,15 @@
...
-from typing import Dict, Iterable, List, Optional, Tuple
+import warnings
+from typing import Dict, Iterable, List, Optional, Tuple, Union
 
 import igraph
 import networkx as nx
 import numpy as np
 import pandas as pd
+import polars as pl
+from networkx.algorithms import bipartite
```
ambarrio, line 17: I would like to make more explicit that `bipartite` is  coming from `networkx` so it is easier to separate igraph from networkx

src/pixelator/graph/backends/implementations.py
```diff
@@ -6,12 +6,15 @@
 from __future__ import annotations
 
 import logging
-from typing import Dict, Iterable, List, Optional, Tuple
+import warnings
+from typing import Dict, Iterable, List, Optional, Tuple, Union
 
 import igraph
```
ambarrio, line 12: To separate better, should we move this imports within the classes? Or the classes on different files?

src/pixelator/graph/backends/implementations.py
```diff
@@ -456,17 +654,36 @@ def from_edgelist(
         :rtype: NetworkXGraphBackend
         :raises: AssertionError when the input edge list is not valid
         """
-        raise NotImplementedError()
+        # TODO If we could change the signature here to work with lazy frames
```
ambarrio, line 657: pl.LazyFrame all the way?

src/pixelator/graph/backends/implementations.py
```diff
@@ -456,17 +654,36 @@ def from_edgelist(
...
+        if add_marker_counts:
+            graph = NetworkXGraphBackend._build_graph_with_marker_counts(
+                edgelist, simplify, use_full_bipartite
+            )
+        else:
+            graph = NetworkXGraphBackend._build_plain_graph(
+                edgelist, simplify, use_full_bipartite
+            )
+
+        # TODO igraph uses integer indexing. This converts the igraph graph to using the
```
ambarrio, line 671: igraph graph -> networkx graph

src/pixelator/graph/backends/implementations.py
```diff
@@ -428,9 +517,118 @@ def __init__(
...
+    def _build_plain_graph(
+        edgelist: pl.LazyFrame, simplify: bool, use_full_bipartite: bool
+    ) -> Union[nx.Graph, nx.MultiGraph]:
+        graph = NetworkXGraphBackend._build_plain_graph_from_edgelist(
+            edgelist.select(["upia", "upib", "umi"]),
+            create_using=nx.Graph if simplify else nx.MultiGraph,
+        )
+        a_nodes = set(edgelist.select(["upia"]).unique().collect()["upia"].to_list())
+        NetworkXGraphBackend._add_node_attributes(graph, a_nodes)
+        if use_full_bipartite:
```
ambarrio, line 625: If we are not doing A-node projection, we don't need to do L623 and 624, right? Should we do an else here?

src/pixelator/graph/backends/implementations.py
```diff
@@ -428,9 +517,118 @@ def __init__(
...
+    def _build_graph_with_marker_counts(
+        edgelist: pl.LazyFrame, simplify: bool, use_full_bipartite: bool
+    ) -> Union[nx.Graph, nx.MultiGraph]:
+        graph = NetworkXGraphBackend._build_graph_with_node_counts_from_edgelist(
+            edgelist,
+            create_using=nx.Graph if simplify else nx.MultiGraph,
+        )
+        a_nodes = set(edgelist.select(["upia"]).unique().collect()["upia"].to_list())
+        NetworkXGraphBackend._add_node_attributes(graph, a_nodes)
+        if use_full_bipartite:
```
ambarrio, line 611: Same here as below, return the graph if we don't need to compute A-node projection

src/pixelator/graph/backends/implementations.py
```diff
@@ -428,9 +517,118 @@ def __init__(
...
+        nx.set_node_attributes(graph, node_names, "name")
+        nx.set_node_attributes(graph, pixel_type, "pixel_type")
+        nx.set_node_attributes(graph, type_, "type")
+
+    @staticmethod
+    def _project_on_a_nodes(
+        graph: Union[nx.Graph, nx.MultiGraph], a_nodes: set[str]
+    ) -> Union[nx.Graph, nx.MultiGraph]:
+        if isinstance(graph, nx.MultiGraph):
+            warnings.warn(
```
ambarrio, line 593: I don't understand the warning. But let's discuss it later.

src/pixelator/graph/backends/implementations.py
```diff
@@ -428,9 +517,118 @@ def __init__(
...
+            for row in df.collect(streaming=True).iter_rows(
+                named=False, buffer_size=1000
+            )
+        )
+        return g
+
+    @staticmethod
+    def _build_graph_with_node_counts_from_edgelist(
+        df: pl.LazyFrame,
+        create_using: bool,
```
ambarrio, line 537: WHat does it mean `create_using`?

src/pixelator/graph/backends/implementations.py
```diff
@@ -428,9 +517,118 @@ def __init__(
...
+        initial_marker_dict = {marker: 0 for marker in unique_markers}
+
+        g: nx.Graph = nx.empty_graph(0, create_using)
+
+        for row in (
+            df.select(["upia", "upib", "marker"])
+            .collect(streaming=True)
+            .iter_rows(named=False, buffer_size=1000)
+        ):
+            # We are duplicating code here, since it gives
```
ambarrio, line 549: What is the part of the code and duplication that is giving this performance boost? Is not entirely clear to me.

src/pixelator/graph/backends/implementations.py
```diff
@@ -428,9 +517,118 @@ def __init__(
...
+
+        return g
+
+    @staticmethod
+    def _add_node_attributes(
+        graph: Union[nx.Graph, nx.MultiGraph], a_nodes: set[str]
+    ) -> None:
+        node_names = {node: node for node in graph.nodes()}
+        pixel_type = {node: "A" if node in a_nodes else "B" for node in graph.nodes()}
+        type_ = {node: node in a_nodes for node in graph.nodes()}
```
ambarrio, line 583: `type_`is only setting A-nodes? Back-underscoring variables is interesting, standard?

ambarrio (commented):
Sending this. We can take a chat online.

src/pixelator/graph/backends/implementations.py
johandahlberg: I do this because `type` is a reserved keyword in python, so then the standard of to add a `_`.

src/pixelator/graph/backends/implementations.py
johandahlberg: Previously there was a function that was used here. @fbdtemme suggested inlineing it: https://github.com/PixelgenTechnologies/pixelator/pull/32#discussion_r1372827702 so now there is a bit more code duplication there.

src/pixelator/graph/backends/implementations.py
johandahlberg: The type annotation here is a bit misleading. I've fixe that now, to make it clear that it relates to which networkx graph class should be used.

tests/graph/test_graph_utils.py
```diff
@@ -63,131 +61,33 @@ def random_sequence(size: int) -> str:
...
-            "median_reads",
-            "mean_upia_degree",
-            "median_upia_degree",
-            "mean_umi_per_upia",
-            "median_umi_per_upia",
-            "upia_per_upib",
-        ]
+    assert_frame_equal(
+        metrics,
+        pd.DataFrame.from_records(
```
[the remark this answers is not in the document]
johandahlberg: Thanks!

DEVELOPERS.md
```diff
@@ -234,3 +234,8 @@ e.g.:
 panel: "human-sc-immunology-spatial-proteomics"
 panel_file: null
 ```
+
+## Pixelator benchmark tests
+
+Pixelator uses `pytest-benchmark` to enable running micro-benchmarks. Normally when running the tests these are disabled.
+You can enabled them by running `pytest --benchmark-enable tests/`.
```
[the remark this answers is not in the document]
johandahlberg: When you enable it it gives you a benchmark report where you use the `benchmark` fixture. They can look something like this. So it's been very useful when trying to get the igraph and networkx performances on par.

```
---------------------------------------------------------------------------- benchmark 'test_build_graph_a_node_projected_benchmark': 2 tests ---------------------------------------------------------------------------
Name (time in ms)                                             Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_build_graph_a_node_projected_benchmark[igraph]       13.1020 (1.0)      16.9474 (1.0)      14.4373 (1.0)      0.7797 (1.0)      14.3642 (1.0)      0.7355 (1.0)          13;3  69.2648 (1.0)          59           1
test_build_graph_a_node_projected_benchmark[networkx]     23.1980 (1.77)     32.5580 (1.92)     26.3172 (1.82)     2.6663 (3.42)     25.2632 (1.76)     3.1088 (4.23)          8;3  37.9979 (0.55)         36           1
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

--------------------------------------------------------------------------- benchmark 'test_build_graph_full_bipartite_do_not_add_marker_counts_benchmark': 2 tests ----------------------------------------------------------------------------
Name (time in ms)                                                                    Min                Max               Mean            StdDev             Median               IQR            Outliers      OPS            Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_build_graph_full_bipartite_do_not_add_marker_counts_benchmark[igraph]        9.0304 (1.0)      16.8877 (1.17)     10.0981 (1.0)      1.5574 (2.11)      9.5493 (1.0)      0.6434 (1.0)          9;11  99.0286 (1.0)          95           1
test_build_graph_full_bipartite_do_not_add_marker_counts_benchmark[networkx]     10.0283 (1.11)     14.4479 (1.0)      11.3390 (1.12)     0.7372 (1.0)      11.2656 (1.18)     0.7728 (1.20)         19;3  88.1911 (0.89)         83           1
```

tests/graph/test_graph.py
```diff
@@ -61,124 +54,166 @@ def create_simple_edge_list_from_graph(
...
-    output_dir.mkdir()
-    yield output_dir
+@pytest.fixture
+def enable_backend(request):
+    previous_environment = os.environ
+    if request.param == "networkx":
+        new_environment = previous_environment.copy()
+        new_environment["ENABLE_NETWORKX_BACKEND"] = True
+        os.environ = new_environment
+    yield
```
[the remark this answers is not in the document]
johandahlberg: You can use `yield` in fixture to have a setup/teardown style test. I.e. first everything above the `yield` statement runs, in this case configuring the environment. `yield` will then yield to the test, i.e. let it run. Then everything below the yield statement will run, in this case restoring the environment.

tests/graph/test_graph.py
johandahlberg: I can show you, but essentially it will setup a microbench mark of the code that is called with `benchmark`. You can find more docs here: https://pytest-benchmark.readthedocs.io/en/stable/

src/pixelator/pixeldataset.py
```diff
@@ -709,14 +709,10 @@ def graph(
         :raises: KeyError if the provided `component_id` is not found in the edgelist
         """
         if component_id:
-            potential_component = _enforce_edgelist_types(
-                (
-                    self.edgelist_lazy.filter(pl.col("component") == component_id)
-                    .collect()
-                    .to_pandas()
-                )
+            potential_component = self.edgelist_lazy.filter(
```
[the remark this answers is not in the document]
johandahlberg: I dropped it since it requires us to materialize the data. It's also not as critical when working on the lazy edge list since, we never bring all of it into memory at the same time (and the reduced memory requirement is the main point of enforcing the types).

src/pixelator/graph/backends/implementations.py
johandahlberg: I forgot to remove this comment, I did implement it like this. I'll remove the comment.

src/pixelator/graph/backends/implementations.py
johandahlberg: I don't think we should import inside the classes, since the recommendation in python is to import at the top of the file so that all imports are resolved at import time. But, we can split this into separate files for igraph and networkx. I suggest we do that in a separate PR though.

src/pixelator/graph/backends/implementations.py
```diff
@@ -428,9 +517,118 @@ def __init__(
...
+        nx.set_node_attributes(graph, pixel_type, "pixel_type")
+        nx.set_node_attributes(graph, type_, "type")
+
+    @staticmethod
+    def _project_on_a_nodes(
+        graph: Union[nx.Graph, nx.MultiGraph], a_nodes: set[str]
+    ) -> Union[nx.Graph, nx.MultiGraph]:
+        if isinstance(graph, nx.MultiGraph):
+            warnings.warn(
+                "Using `use_full_bipartite=True` together with `simplify=False` "
```
ambarrio, line 594: True -> False

src/pixelator/graph/backends/implementations.py
ambarrio: This needs to be in place as it is needed for plotting. It is also tested as is.

ambarrio (approved):
Sending this. We can take a chat online.

Great discussion!

PR #34, opened 2023-10-26
Breaking pixeldataset.py up into separate files

johandahlberg:
## Description

This separate the huge `pixeldataset.py` file into separate smaller files. I suggest focusing on if you think that the module names (and associated documentation) make sense when reviewing this, since nothing else has really been changed.

Fixes: EXE-1078

## Type of change

Purely refactoring.

## How Has This Been Tested?

Ran unit tests locally.


src/pixelator/annotate/__init__.py
```diff
@@ -1,6 +1,4 @@
-"""
-This module contains functions for the filtering and annotation of
-pixel data in edge list format.
+"""Functions for filtering and annotating of pixel data in edge list format.
```
ambarrio, line 1: pixel -> MPX

src/pixelator/pixeldataset/__init__.py
```diff
@@ -0,0 +1,400 @@
...
+
+    @property
+    def adata(self) -> AnnData:
+        """Get the AnnData object."""
+        return self._backend.adata
+
+    @adata.setter
+    def adata(self, value: AnnData) -> None:
+        """Set the AnnData object."""
+        self._backend.adata = value
```
ambarrio, line 150: Nice add-on with codecov!

src/pixelator/pixeldataset/__init__.py
```diff
@@ -0,0 +1,400 @@
...
+        """Get the metadata dictionary."""
+        metadata = self._backend.metadata
+        if not metadata:
+            return {}
+        return metadata
+
+    @metadata.setter
+    def metadata(self, value: Dict) -> None:
+        """Set the metadata object."""
+        self._backend.metadata = value
```
ambarrio, line 205: It seems clear we are not testing the setters.

src/pixelator/pixeldataset/aggregation.py
```diff
@@ -0,0 +1,157 @@
...
+
+def simple_aggregate(
+    sample_names: List[str],
+    datasets: List[PixelDataset],
+    ignore_edgelists: bool = False,
+) -> PixelDataset:
+    """Aggregate samples in a simple way (see caveats).
+
+    Aggregating samples in a simplistic fashion. This function should only
+    be used if the dataset you merge have been generated with the same panel.
```
ambarrio, line 68: Is there anyway we could assert for this in the function? I think it should be possible as we store the panel and could easily check the metadata or the targets.

ambarrio (commented):
See my comments and the codecov reports.

src/pixelator/pixeldataset/__init__.py
johandahlberg: Yeah. However, I suggest we stick to only the refactoring in this PR, and go on with improving test coverage later. I think it's easier when large scale moves (as this one) are not conflated with adding/removing code.

src/pixelator/pixeldataset/aggregation.py
johandahlberg: Right now we require that the antibodies are the same, but we probably want to do a more advanced aggregation here. Perhaps you want to add a ticket about that and we can follow up on that?

fbdtemme (approved):

PR #43, opened 2023-11-22
Drop py38 and py39 support

fbdtemme:
## Description

Drop python 3.8 and 3.9 since those are not supported anymore by some dependencies.

Fixes: [EXE-1233](https://linear.app/pixelgen-technologies/issue/EXE-1233/drop-py38-and-py39)

## Type of change


tests/graph/test_graph_utils.py
```diff
@@ -76,7 +77,7 @@ def test_components_metrics(full_graph_edgelist: pd.DataFrame):
                     "upia": 50,
                     "upib": 50,
                     "umi": 1860,
-                    "reads": 2500,
+                    "reads": np.uint64(2500),
```
ambarrio, line 80: I understand the `np` changes are for the anndata warnings, right?

ambarrio (approved):
LGTM! Just small comment.

tests/graph/test_graph_utils.py
fbdtemme: A few test cases failed due to different data types after updating dependencies. This is just to make sure the dataframe we compare against has the same column types.

johandahlberg (approved):
lgtm

PR #45, opened 2023-11-27
Feature/exe 1196 refactor implementations

johandahlberg:
## Description

Refactor the implementations model into separate parts to make modules smaller.

Fixes: EXE-1196

## Type of change

Mainly refactoring

## How Has This Been Tested?

Tested with the unit test suite.

### DIFF
--- a/src/pixelator/graph/backends/implementations/__init__.py
+++ b/src/pixelator/graph/backends/implementations/__init__.py
@@ -32,7 +32,7 @@ def graph_backend(
     :raises ValueError: when `graph_backend_class` is not recognized
     """
     # TODO Later on we could use this as an entry point for loading
-    # graph backend from plugins if we would like.
+    # a graph backend from plugins if we would like.
 
     def _load_nx():
         logger.debug("Setting up a networkx based backend")


src/pixelator/graph/backends/implementations/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+        logger.debug("Setting up an igraph based backend")
+        return IgraphGraphBackend
+
+    if not graph_backend_class:
+        if str(os.environ.get("ENABLE_NETWORKX_BACKEND", False)) in (
+            "True",
+            "true",
+            "1",
+        ):
+            return _load_nx()
```
fbdtemme, line 51: I try to avoid having these kinds of variables that include one of the options and prefer something like PIXELATOR_GRAPH_BACKEND="networkx"  (or maybe PXL_GRAPH_BACKEND). The common prefix for all environment variables helps to quickly find the ones related to pixelator when debugging environments and also helps to reduce conflicts (probably not an issue here but sometimes the names can get quite general). And the string variable is of course to allow other backends in the future without adding a separate toggle.

src/pixelator/graph/backends/implementations/_igraph.py
```diff
@@ -0,0 +1,536 @@
...
+                n_iterations=n_iterations,
+                beta=beta,
+                **kwargs,
+            ),
+            graph=self._raw,
+        )
+
+    def layout_coordinates(
+        self,
+        layout_algorithm: str = "fruchterman_reingold",
```
fbdtemme, line 221: Maybe we can encode the options in a `typing.Literal` here?

fbdtemme (approved):
Just made 2 small suggestions, but feel free to ignore them or maybe take up later since they are not about the refactoring per se.
It all looks very clean!

src/pixelator/graph/backends/implementations/_igraph.py
johandahlberg: I suggest sticking with this for now. But I will revisit this later as I work on the layout stuff in networkx.

src/pixelator/graph/backends/implementations/__init__.py
johandahlberg: I think that is an excellent suggestion. I suggest sticking with this for now, and then I will change this in different PR to try to stick mostly to the refactoring here.

PR #53, opened 2023-12-18
Move all tests to use networkx

johandahlberg:
## Description

Move all tests to using networkx as their backend.

Fixes: EXE-1263

## Type of change

## How Has This Been Tested?

This has been tested by running the unit test suite.

## PR checklist:

### DIFF
--- a/src/pixelator/graph/backends/implementations/_networkx.py
+++ b/src/pixelator/graph/backends/implementations/_networkx.py
@@ -250,7 +250,8 @@ class NetworkXGraphBackend(GraphBackend):
         """Get a sequence of the vertices in the Graph instance."""
         return NetworkxBasedVertexSequence(
             vertices=[
-                NetworkxBasedVertex(v[0], v[1]) for v in self._raw.nodes(data=True)
+                NetworkxBasedVertex(v[0], v[1], self._raw)
+                for v in self._raw.nodes(data=True)
             ]
         )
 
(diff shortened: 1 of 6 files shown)


tests/graph/igraph/test_tools.py
ambarrio, line 1: This whole thing is in a module called `graph/igraph/test_tools.py`. The idea was to replicate the nx stuff on `graph/networkx/test_tools.py` to use alongside. And if it works, just remove the `igraph`folder. Otherwise, I think it looks good.

tests/analysis/polarization/test_polarization.py
```diff
@@ -36,14 +36,14 @@ def test_polarization(enable_backend, full_graph_edgelist: pd.DataFrame):
     assert_frame_equal(scores, expected)
 
 
-# TODO Add test here that tests networkx backend
 def test_polarization_with_differentially_polarized_markers():
     # Set seed to get same graph every time
     graph = create_randomly_connected_bipartite_graph(
-        n1=50, n2=100, p=0.1, random_seed=1477
+        n1=50, n2=100, p=0.1, random_seed=2
```
ambarrio, line 42: WHy the difference in seed here? Is it important?

tests/conftest.py
ambarrio, line 1: Would this work both in igraph and nx? Let's talk about it, but maybe it would be interesting to have the same way of accessing both

ambarrio (commented):
A few spurious comments and a couple of decisions to take. We can talk over them.

src/pixelator/graph/backends/implementations/_networkx.py
```diff
@@ -488,6 +490,17 @@ def __setitem__(self, attr: str, value: Any) -> None:
         """Set the attr of the vertex."""
         self._data[attr] = value
 
+    def neighbors(self) -> VertexSequence:
+        """Get the neighbors of the vertex."""
+        neighbor_vertices = set(self._graph.neighbors(self._index))
+
+        def generate_neighbors():
+            for node_idx, data in self._graph.nodes(data=True):
```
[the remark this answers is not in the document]
johandahlberg: We do get it directly from the graph, see L495. However, in order to extract the associated data to populate the `NetworkxBasedVertex` data of the need to iterate over the nodes. This is as far as I can see now no way around this due to the way networkx works (short of keeping all copy of all the node data outside of the graph). Unfortunately, I see no way around it here.

tests/conftest.py
johandahlberg: I think we could make it work with both without to much extra work if you want that. I just followed the simplest path here to make the implementation as easy as possible.

tests/graph/igraph/test_tools.py
johandahlberg: Ah! Yes. I had actually missed that this was placed in the igraph module. I just switched it all to use networkx instead. We can duplicate the same functionality and keep both, or just rename one depending on what you prefer.

tests/analysis/polarization/test_polarization.py
johandahlberg: Using 1477 as a random seed with networkx here created a graph with two components, so I changed the random seed until I found one that generated a single connected component with the same settings for n1, n2, and p.

ambarrio (approved):
LGTM 👍

PR #54, opened 2023-12-22
Formatting fixes, and exposing `raw` as public property

johandahlberg:
## Description

- Fixes incorrect formatting in analysis command cli docs.
- Exposes that `raw` property of the `Graph` object as a public property.

## Type of change

With unit tests and by manually looking at the docs.


ambarrio (approved):
LGTM

PR #57, opened 2024-01-09
Feature/exe 1103 remove igraph as dependency

johandahlberg:
## Description

This PR removes igraph, leiden, and louvain as dependencies of pixelator.

It also adds a few recommended vscode extensions to the repo to make it easier for new devs to get up and running with the project.

Fixes: EXE-1103

## Type of change

## How Has This Been Tested?

I have run the unit test suite.


tests/graph/test_graph.py
```diff
@@ -185,12 +184,12 @@ def test_build_graph_a_node_projected_benchmark(
...
-@pytest.mark.parametrize("enable_backend", ["igraph", "networkx"], indirect=True)
+@pytest.mark.parametrize("enable_backend", ["networkx"], indirect=True)
 def test_build_graph_a_node_projected_without_simplifying(
     enable_backend,
     full_graph_edgelist: pd.DataFrame,
 ):
-    def _test():
+    with pytest.warns(UserWarning):
         # The A-node projection disregards any multiedges, so running it with
         # or with out simplification should yield the same result
```
ambarrio, line 194: with out -> without

src/pixelator/annotate/__init__.py
```diff
@@ -229,6 +232,15 @@ def annotate_components(
         json.dump(edgelist_metrics(edgelist), outfile, default=np_encoder)
 
 
+def _cluster_components_using_leiden(adata: AnnData) -> None:
```
ambarrio, line 235: should we expose a `resolution` parameter here that defaults to 1.0? It is the only thing I can think of. But I think it works as is.

src/pixelator/annotate/__init__.py
```diff
@@ -229,6 +232,15 @@ def annotate_components(
...
+    """Carry out a leiden clustering on the components."""
+    g = nx.from_scipy_sparse_array(adata.obsp["connectivities"])
+    partitions = leiden(g, resolution=1.0)
+    partitions_df = pd.DataFrame.from_dict(partitions, orient="index").sort_index()
```
ambarrio, line 239: Just curious about why the need of `sort_index` here?

ambarrio (commented):
Just few word checks and questions

src/pixelator/annotate/__init__.py
johandahlberg: Sure, I'll add that here.

src/pixelator/annotate/__init__.py
johandahlberg: The reason is that the partitions are returned as a dict from `leiden`, and that when creating a data frame from them there is no guarantee of the order they end up in, so I sort them so I can then just assigned them directly as a column to the `adata.obs`.

ambarrio (approved):
LGTM

PR #68, opened 2024-01-22
Feature/update qc report 0.11.0

fbdtemme:
## Description

Update the qc report template to v0.10.0

This brings in the improved performance and smaller bundle size changes from https://github.com/PixelgenTechnologies/pixelator-qc-report/pull/32.

The `antibody-percentages` input is now also compressed.


johandahlberg (approved):
This needs an updated to the change-log, but otherwise I think this looks good to go.

PR #69, opened 2024-01-24
Fix deflated counts in edgelist after collapse

fbdtemme:
Fix an issue were read counts in the edgelist are deflated because the counts of collapsed reads are not taken into account.
Only the reads of unique matches to the corrected read were reported and not the reads of other molecules that have mismatches.

Fixes: EXE-1250

## Type of change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 * Performance improvements and reduced bundle size in QC report.
+* Fix deflated counts in the edgelist after collapse.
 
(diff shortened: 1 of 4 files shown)


ambarrio:
I think we should merge this first and then, look at the other metrics after. Let me check quickly your PR.

ambarrio (approved):
Great work!

fbdtemme:
I will run the samplesheet_full test profile with nf-core/pixelator to compare the impact on the metrics before merging.

ambarrio (approved):

johandahlberg (approved):
Nice work!

One thing I noticed was the change of type-hinting with `list` rather than `typing.List` - I see that seems to be supported from python 3.8 and up. It will take some getting used to for me. :D

PR #73, opened 2024-01-29
Fix bug in calculation of discarded UMIs

stefanppetkov:
## Description

Calculate discarded UMIs based on data from collapse, graph, annotate and cell calling as opposed to just using discarded aggregates.

Fixes TECH-29

## Type of change

## How Has This Been Tested?

Generate QC report with updated code by running:

`pixelator single-cell report`

## PR checklist:

- [x] Update [CHANGELOG.md](../CHANGELOG.md)

### DIFF
--- a/src/pixelator/report/__init__.py
+++ b/src/pixelator/report/__init__.py
@@ -674,8 +674,15 @@ def create_dynamic_report(
         "fraction_predicted_cell_type_unknown": None,
     }
 
+    umi_counts = {
+        "after_collapse": summary_collapse["output_umi"],
+        "after_graph": summary_graph["umi"],
+        "after_annotate": summary_annotate["umi"],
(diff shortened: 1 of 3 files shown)


src/pixelator/report/__init__.py
```diff
@@ -674,8 +674,15 @@ def create_dynamic_report(
         "fraction_predicted_cell_type_unknown": None,
     }
 
+    umi_counts = {
+        "after_collapse": summary_collapse["output_umi"],
+        "after_graph": summary_graph["umi"],
+        "after_annotate": summary_annotate["umi"],
```
johandahlberg, line 680: I suggest removing `after_graph` and `after_annotate` since we don't use them.

johandahlberg (approved):
Small suggestion. Fix that, and then I think this is good to merge. Nice catch!

PR #75, opened 2024-01-30
Remove multi-sample processing

fbdtemme:
## Description

Remove parallel processing of multiple samples in a single command. 
This feature is better handled by using the nextflow pipeline.

The only parallel handling remaining on the input file level is in collapse to deal with the demuxed fastq files.

The `amplicon` subcommand interface is changed: 
The also removes `--input1-pattern` and `--input2-pattern` since these have become redundant.
Add a --sample-name flag to use a custom output filename instead of extracting a sample name from the input fastq files.

Fixes: EXE-1341

## Type of change

### DIFF
--- a/src/pixelator/utils.py
+++ b/src/pixelator/utils.py
@@ -17,7 +17,17 @@ import typing
 from concurrent import futures
 from functools import wraps
 from pathlib import Path, PurePath
-from typing import Any, Dict, List, Optional, Sequence, Set, TYPE_CHECKING, Union
+from typing import (
+    Any,
+    Dict,
+    List,
+    Optional,
+    Sequence,
+    Set,
+    TYPE_CHECKING,
+    Union,
+    Literal,
+)
 
 import click
 import numpy as np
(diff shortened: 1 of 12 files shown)


tests/utils/test_utils.py
```diff
@@ -146,3 +148,56 @@ def test_multiprocess_logging(verbose):
...
+    assert get_read_sample_name("sample1_R2.fq.gz") == "sample1"
+    assert get_read_sample_name("sample1_r1.fq.gz") == "sample1"
+    assert get_read_sample_name("sample1_r2.fq.gz") == "sample1"
+    assert get_read_sample_name("sample1.r1.fq.gz") == "sample1"
+    assert get_read_sample_name("sample1.r2.fq.gz") == "sample1"
+    assert get_read_sample_name("sample1.R1.fq.gz") == "sample1"
+    assert get_read_sample_name("sample1.R2.fq.gz") == "sample1"
+
+    assert get_read_sample_name("sample1_R1_L001.fq.gz") == "sample1_L001"
+    assert get_read_sample_name("sample1_R2_L001.fq.gz") == "sample1_L001"
```
ambarrio, line 173: I am not sure if I like to keep the lane name here. And I know this is a  `get_read_sample_name` change, but I am unsure what value brings keeping lanes in sample names. Maybe is good somehow to keep it as a reminiscent of the sequencing design.

src/pixelator/utils.py
```diff
@@ -9,14 +9,25 @@
...
+from typing import (
+    Any,
+    Dict,
+    List,
+    Optional,
+    Sequence,
+    Set,
+    TYPE_CHECKING,
+    Union,
+    Literal,
```
ambarrio, line 29: Should we just put `Literal` in lexicographic order? I know, nitty-picky.

src/pixelator/utils.py
```diff
@@ -360,3 +371,60 @@ def get_process_pool_executor(ctx, **kwargs):
...
+def get_read_sample_name(read: str) -> str:
+    """Extract the sample name from a read file.
+
+    Strip fq.gz or fastq.gz extension and remove R1/R2 suffixes.
+    Supported R1 R2 identifieds are:
+
+    _R1,_R2 | _r1, _r2 | _1, _2 | .R1, .R2 | .r1, .r2
+
+    :param read: filename of a fastq read file
+    :return str: sample name
```
ambarrio, line 385: return vs returns?  Are both supported in sphinx and autodoc?

ambarrio (approved):
LGTM, see my comments. Mostly curiosity or stupid details.

src/pixelator/utils.py
```diff
@@ -51,7 +62,7 @@ def build_barcodes_file(
     :param anchored: make the sequences anchored if True
     :param rev_complement: reverse complement the sequence column
                            if True
-    :returns: a path to the barcodes file
+    :returns str: a path to the barcodes file
```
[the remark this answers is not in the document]
johandahlberg: Yeah, that is a good point.

tests/utils/test_utils.py
johandahlberg: My 2 cents is not to include the lane number in the sample name, since I suppose the same sample might be run across multiple lanes?

johandahlberg:
> > find a log statement indicating
> 
> Warning log level is also printed on the console so it should be quite obvious. But anyways, I have changed it to errors by default and warnings when `--skip-input-checks` is given.

Good I wasn't sure it was. But, I think this might cause folks less trouble in the long run.

tests/utils/test_utils.py
ambarrio: That's a good point. However, when running across multiple lanes, we have this feature in the samplesheets that using the same ID across multiple rows with different libraries/indexes, you can pool reads together by concatenation.

johandahlberg: Aha. Cool. But then I think we really should just keep the actual sample name, stripping the lane part.

ambarrio: Ok, it shouldn't be too difficult to implement, right @fbdtemme ?

fbdtemme: I would not remove anything except the R1/R2 identifier to create the default output sample names.
I think if we remove lane suffixes we create the impression that pixelator somehow "handles" this in a way.

I would prefer not to do anything smart with the name generation since that will always be quite brittle with regexes and needing to handle different naming conventions etc...

You can always use the new `--sample-name` option to override the default or just rename your input files to something sensible in the first place to get nicer output names by default.

src/pixelator/utils.py
fbdtemme: Yes, they are equivalent.

PR #76, opened 2024-01-31
The edgelist will now be collected only once in the parent functions …

ptajvar:
## Description

A small change in the implementation of the pixelator Graph that does not affect the usage of functions. The collect() function from polars that converts a LazyFrame to a DataFrame in the memory is the main speed bottleneck of Graph.from_edgelist method; the code is as a result modified to only call it once for each graph construction and pass on the DataFrame instead of the LazyFrame to the subsequent functions.

Fixes # EXE-1318

## Type of change

## How Has This Been Tested?

Graph.from_edgelist method with all different combinations of simplify, use_full_bipartite, and add_marker_counts. Unit tests and the small workflow test are run. 

## PR checklist:

- [x] My code follows the style guidelines of this project
- [x] Make sure your code lints, is well-formatted and type checks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Fix deflated counts in the edgelist after collapse.
 * Improved console output in verbose mode.
 * Improved logging from multiprocessing jobs.
+* Improved runtime for graph synthesis.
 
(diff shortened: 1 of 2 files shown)


johandahlberg (approved):
This looks great! I also did some additional tests on this running the `graph` step on a sample locally just to check that it didn't have an adverse impact on the memory usage, and it was more or less the same.

Add a note to the CHANGELOG and merge at your convenience.

PR #85, opened 2024-02-06
Validate pydocstyle rules with Ruff

johandahlberg:
## Description

Remove pydocstyle rules to be validated by ruff.

Fixes: EXE-1364

## Type of change

This only changes linting rules. No impact on actual production code.


fbdtemme (approved):
Heroic effort fixing all the docstrings :+1:

PR #86, opened 2024-02-07
Add __repr__ method to graph

johandahlberg:
## Description

Adds a `__repr__` method to the `Graph` class to give some nice output in e.g. jupyter notebooks.

Fixes: EXE-915

## Type of change

## How Has This Been Tested?

Unit tests and by trying it in ipython.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -9,4 +9,4 @@ arg-type-hints-in-signature = True
 arg-type-hints-in-docstring = False
 # Ignore documentation linting in tests
 per-file-ignores =
-    tests/**: DOC
+    tests/**: DOC, D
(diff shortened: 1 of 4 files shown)


fbdtemme (approved):

.flake8
```diff
@@ -9,4 +9,4 @@ arg-type-hints-in-signature = True
 arg-type-hints-in-docstring = False
 # Ignore documentation linting in tests
 per-file-ignores =
-    tests/**: DOC
+    tests/**: DOC, D
```
ambarrio, line 12: What is `D` here?

johandahlberg: `D` is for `pydocstyle` rules. I think I accidentally removed ignoring it when was moving pydocstyle validation to ruff instead of flake8.

PR #92, opened 2024-02-20
Add nf-core-pixelator integration workflow

fbdtemme:
## Description
 
Add a workflow to check [`@PixelgenTechnologies/nf-core-pixelator:pixelator-next`](https://github.com/PixelgenTechnologies/nf-core-pixelator/tree/pixelator-next)  compatibility.

Example workflow run:

https://github.com/PixelgenTechnologies/pixelator/actions/runs/7974606851/job/21770903461

For now since the workflow is not yet on the main branch it is not possibly to trigger the workflow run from the UI.
You can launch it using the gh cli instead:

```shell
gh workflow run nf-core-pixelator-tests.yml --ref=feature/nf-pixelator-integration-workflow
```


.github/workflows/checks.yml
```diff
@@ -72,29 +72,26 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Check code lints with Black
-        uses: psf/black@stable
+      - name: ruff formatter
+        uses: chartboost/ruff-action@v1
         with:
-          version: 23.12.0
+          args: "format --check"
```
johandahlberg, line 78: Good to switch out black. I think it would be neat if we could also drop flake8 completely in favor of ruff I think.

.github/workflows/nf-core-pixelator-tests.yml
```diff
@@ -0,0 +1,93 @@
...
+          version: "latest-stable"
+
+      - name: Cache Conda env
+        uses: actions/cache@v3
+        with:
+          path: ${{ env.CONDA }}/envs
+          key:
+            conda-${{ runner.os }}--${{ runner.arch }}--${{ env.CACHE_NUMBER }}--${{ hashFiles('environment.yml') }}
+        env:
+          # Increase this value to reset cache if etc/example-environment.yml has not changed
```
johandahlberg, line 73: Not sure if this comments makes sense in this context since I guess we don't have an `etc/example-environment.yml` file?

johandahlberg (approved):
This looks very good I think. One thing that I think we should perhaps consider is adding default edit config files here, at least for vscode which I guess is the most common choice to make sure that folks get correct formatting/linting directly in the editor.

ambarrio (approved):
A few questions

ambarrio (commented):
A few questions

ambarrio (commented):
A few questions

ambarrio (commented):
Questions to discuss

.github/workflows/nf-core-pixelator-tests.yml
```diff
@@ -0,0 +1,93 @@
...
+          use-mamba: true
+          environment-file: environment.yml
+
+      - name: Install pixelator
+        run: pip install -e .
+
+      - name: Run nf-core/pixelator default profile with current pixelator
+        env:
+          PIPELINE_BRANCH: pixelator-next
+          PIPELINE_SOURCE_DIR: ./nf-core-pixelator
```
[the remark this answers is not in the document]
fbdtemme: The location of the pipeline code can be set with this ENV variable to deal with different situations. You can either let the taskfile clone the repo or clone the repo yourself and provide the path to it. In the CI worfklow we just need to set the path since we already cloned it.

.github/workflows/checks.yml
fbdtemme: Agreed. I have removed all flake8 setup now.

ambarrio (approved):
LGTM

PR #93, opened 2024-02-20
Feature/exe 1328 add layouts to pixeldataset tidy

johandahlberg:
## Description

This does a bunch of different things (sorry in advance about that).

### EXE-1328: make it possible to add pre-computed layouts to pxl files

This adds the capability for pixel files to hold pre-computed layouts. This is useful since it allows us to interactively visualize components in downstream analysis without having to do the relatively computationally expensive layout calculations on the fly.

### EXE-1414: a bug that caused the the same file to be written multiple time to the pxl file

When writing a file with the same file name multiple times to the pxl file, this would appear multiple times in the file. Since there is currently (as far as I understand) no way to delete files from a zip-archive in python pixelator will now require a completely fresh file to be written, unless the user explicitly overwrites the old file.

### EXE-1416: add a base exception for pixelator

Add a base exception that we should start using to build exception hierarchies in pixelator to allow us better error handling in the future.

### Makes PMDS use sparse matrices

An issue in the types of the sparse matrix indices meant that we were using dense matrix representation when computing pmds layouts. I found a work-around for this that I have implemented as part of this PR.

### Split `test_pixeldataset.py` into multiple files

This test file was testing multiple things and was getting large enough to be unwieldy. I split it up into multiple files.

## Type of change

## How Has This Been Tested?

With the unit tests, and by running the prototype code included here (but not finished) that can generate layouts, and then trying those out in a notebook.

## PR checklist:

### DIFF
--- a/src/pixelator/pixeldataset/precomputed_layouts.py
+++ b/src/pixelator/pixeldataset/precomputed_layouts.py
@@ -1,0 +1,350 @@
+"""Pre-computed layouts for components.
+
+This module contains the functionality to generate and work with pre-computed
+graph layouts. The primary purpose of this is to allow components to be
+visualized quickly in downstream analysis, since layout computation
+is a relatively computationally expensive operation.
+
+Copyright (c) 2023 Pixelgen Technologies AB.
+"""
+
+from __future__ import annotations
+
+from multiprocessing import get_context
+from typing import TYPE_CHECKING, Iterable, Optional
+
+import pandas as pd
+import polars as pl
+
+from pixelator.graph import Graph
+from pixelator.exceptions import PixelatorBaseException
+from pixelator.marks import experimental
+from pixelator.utils import batched
+
+if TYPE_CHECKING:
+    from pixelator.pixeldataset import PixelDataset
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class PreComputedLayoutsEmpty(PixelatorBaseException):
+    """Raised when trying to access an empty PreComputedLayouts instance."""
+
+
+class PreComputedLayouts:
+    """Pre-computed layouts for a set of graphs, typically per component."""
+
+    DEFAULT_PARTITIONING = [
+        "graph_projection",
+        "layout",
+        "component",
+    ]
+
+    def __init__(
+        self,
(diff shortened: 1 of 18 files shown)


src/pixelator/pixeldataset/precomputed_layouts.py
```diff
@@ -0,0 +1,350 @@
...
+        serializing it to disk, for the cases when this is applicable.
+        For example when writing hive style parquet files.
+        """
+        return self._partitioning
+
+    @property
+    def df(self) -> pd.DataFrame:
+        """Return the layouts as a pandas DataFrame."""
+        return self.lazy.collect().to_pandas()
+
```
fbdtemme, line 94: Small nitpick but I think this would be better renamed to a `to_pandas()` function to make it more clear that this doing some work and data is copied. Alternatively, it could be a cached_property ?

src/pixelator/utils.py
```diff
@@ -116,6 +116,18 @@ def flatten(list_of_collections: List[Union[List, Set]]) -> List:
...
 
+def batched(iterable, n):
+    """Batch data into tuples of length n. The last batch may be shorter.
+
+    Taken from python itertools recipes.
+    """
+    if n < 1:
+        raise ValueError("n must be at least one")
+    it = iter(iterable)
+    while batch := tuple(itertools.islice(it, n)):
```
fbdtemme, line 127: The walrus operator! I haven't found much use for it but it works very nice here.

fbdtemme (approved):

src/pixelator/pixeldataset/precomputed_layouts.py
johandahlberg: I was inspired by this: https://anndata.readthedocs.io/en/latest/generated/anndata.AnnData.to_df.html My thinking was that since `anndata` objects has that type of function if would be good for the user to be able to do something similar here.

However, I agree it would be good to indicate that it actually does some work. Perhaps we should make it a function, and call it `to_df()`, then we would be even more similar to the `anndata` api, and we would also indicate that the user should expect that some work is being done. What do you think?

fbdtemme: Fully agree with making it a `to_df()` function.

PR #96, opened 2024-02-26
Make pixelator use client/server style logging

johandahlberg:
## Description

The previous logging solution would deadlock when creating Processes to use in multiprocessing. Switching to a client/server logging style should fix this.

What we do here is that we create a small TCP server that listens for log records on a socket. This will then output them either to the console or to the log depending on what is available.

Fixes: EXE-1451

## Type of change

## How Has This Been Tested?

With unit tests, and by running the `pixelator singel-cell analysis` command to see that the logs end up in file/console as expected.


fbdtemme (approved):
LGTM :+1:

PR #98, opened 2024-02-29
Fails to write very large pxl files

johandahlberg:
## Description

This PR fixes a problem we have with writing large pxl-files larger than 2Gb. This is related to this issue in https://github.com/fsspec/filesystem_spec/issues/1474

It also fixes some other issues I found:
 - Using pyarrow to write caused us to have smaller row groups in the parquet files which gave us much worse compression rations (files growing up to twice the size)
 - Make sure we use `zstd` compression instead of the default `snappy` for better compression rates
 - A bug that caused aggregation to fail when there are no precomputed layouts in the file

Fixes:
 - EXE-1450

## Type of change

## How Has This Been Tested?

I have run aggregation of a bunch of very large pxl files (without precomputed layouts), and it write an output file larger than 2Gb.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,7 +10,7 @@
         "editor.formatOnSave": true,
         "editor.defaultFormatter": "charliermarsh.ruff",
         "editor.codeActionsOnSave": {
-            "source.organizeImports": true
+            "source.organizeImports": "never"
         }
     },
 }
--- a/src/pixelator/pixeldataset/datastores.py
+++ b/src/pixelator/pixeldataset/datastores.py
@@ -619,6 +619,11 @@ class ZipBasedPixelFileWithParquet(ZipBasedPixelFile):
             where=key,
             filesystem=self._file_system,
             compression=DEFAULT_COMPRESSION,
+            # We want all the data to go into one
+            # parquet row group to allow for maximum compression
+            # when we write the data here.
+            # the `or 1` ensures that it does not raise if
+            # the dataframe is empty.
             row_group_size=len(dataframe) or 1,
         )
 
--- a/src/pixelator/pixeldataset/datastores.py
+++ b/src/pixelator/pixeldataset/datastores.py
@@ -30,7 +30,10 @@ import pyarrow.parquet as pq
 from anndata import AnnData
 from fsspec.implementations.zip import ZipFileSystem
 
+from functools import partial
+
 from pixelator.exceptions import PixelatorBaseException
+
 from pixelator.pixeldataset.precomputed_layouts import PreComputedLayouts
 
 if TYPE_CHECKING:
@@ -295,9 +298,15 @@ class ZipBasedPixelFile(PixelDataStore):
         self.close()
 
     def _setup_file_system(self, mode):
-        self._file_system_handle = ZipFileSystem(
-            fo=self.path, mode=mode, allowZip64=True
-        )
+        files_system = ZipFileSystem(fo=self.path, mode=mode, allowZip64=True)
+
+        # For now we are overwriting the zip open method to force it to
+        # always have force_zip=True, otherwise it won't work for large files
+        def custom_open(f):
+            return partial(f, force_zip64=True)
+
+        files_system.zip.open = custom_open(files_system.zip.open)
+        self._file_system_handle = files_system
         self._current_mode = mode
 
     @property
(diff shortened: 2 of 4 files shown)


.vscode/settings.json
```diff
@@ -10,7 +10,7 @@
         "editor.formatOnSave": true,
         "editor.defaultFormatter": "charliermarsh.ruff",
         "editor.codeActionsOnSave": {
-            "source.organizeImports": true
+            "source.organizeImports": "never"
```
johandahlberg, line 13: Having this on was causing some strange bugs in my vscode where the last line was duplicated every time I saved. So I turned it off.

src/pixelator/pixeldataset/precomputed_layouts.py
```diff
@@ -232,15 +232,20 @@ def data():
...
-    return PreComputedLayouts(
-        pl.concat(data(), rechunk=False),
-        partitioning=["sample"] + PreComputedLayouts.DEFAULT_PARTITIONING,
-    )
+    try:
+        return PreComputedLayouts(
+            pl.concat(data(), rechunk=False),
+            partitioning=["sample"] + PreComputedLayouts.DEFAULT_PARTITIONING,
+        )
+    except ValueError:
```
ambarrio, line 247: Can this be included as one of these pixelator exceptions? What does it mean `ValueError` here?

src/pixelator/pixeldataset/precomputed_layouts.py
```diff
@@ -232,15 +232,20 @@ def data():
...
             yield layout_with_name
 
-    return PreComputedLayouts(
-        pl.concat(data(), rechunk=False),
-        partitioning=["sample"] + PreComputedLayouts.DEFAULT_PARTITIONING,
-    )
+    try:
+        return PreComputedLayouts(
+            pl.concat(data(), rechunk=False),
+            partitioning=["sample"] + PreComputedLayouts.DEFAULT_PARTITIONING,
```
ambarrio, line 245: Is DEFAULT_PARTITIONING going to be only used in PreComputedLayouts? If so, I think this is ok. Otherwise, as a default, should we make this variable private and use a getter?

src/pixelator/pixeldataset/datastores.py
```diff
@@ -590,13 +608,18 @@ def write_dataframe(
...
                     use_threads=False,
+                    file_options=file_options,
                     existing_data_behavior="overwrite_or_ignore",
                 )
             return
 
         self._check_if_writeable(key)
         pq.write_table(
-            pa.Table.from_pandas(dataframe), where=key, filesystem=self._file_system
+            pa.Table.from_pandas(dataframe, preserve_index=False),
```
ambarrio, line 618: I think we were talking about this, but can you remind me why `presenrve_index=False` was key here?

src/pixelator/pixeldataset/datastores.py
```diff
@@ -590,13 +608,18 @@ def write_dataframe(
...
+            where=key,
+            filesystem=self._file_system,
+            compression=DEFAULT_COMPRESSION,
+            row_group_size=len(dataframe) or 1,
```
ambarrio, line 622: MAybe it is not immediate on why is like this. I wonder if we should include a comment with your findings yesterday afternoon

ambarrio (commented):
Great PR! Thank you and I have mostly comments and questions that can be addressed probably quickly.

src/pixelator/pixeldataset/datastores.py
```diff
@@ -579,8 +592,13 @@ def write_dataframe(
         :param key: The key of the dataframe to write.
         :param partitioning: The partitioning to use when writing the dataframe.
         """
+        DEFAULT_COMPRESSION = "zstd"
```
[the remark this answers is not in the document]
johandahlberg: As you say, it's not really reachable from the outer scope anyway, so I think it can stay like this. I made it upper case to mark it as a constant.

src/pixelator/pixeldataset/precomputed_layouts.py
johandahlberg: In this case the `ValueError` is raised by `pl.concat` if there is no data in `data()`, and since we catch it to make sure we return an empty precomputed layout object I don't think we need to convert it into a specific handled exception.

src/pixelator/pixeldataset/precomputed_layouts.py
johandahlberg: My idea was to add it as a constant class variable that can also be used in other parts of the code as needed, and since getters and setters normally deal with instance variables I don't think it's appropriate in this case.

.vscode/settings.json
johandahlberg: Yes.

src/pixelator/pixeldataset/datastores.py
johandahlberg: `preserve_index=False` means that we don't create an additional column for the index from what ever is in the pandas index. See here: https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.from_pandas 

Adding this in this context means that we preserve the behavior we have had in pixelator so far, i.e. no specific index column is stored in the files.

src/pixelator/pixeldataset/datastores.py
johandahlberg: Good point! I will do that.

ambarrio (approved):
Good job!

PR #99, opened 2024-02-29
Add column filtering to iterator/to_df methods

johandahlberg:
## Description

Lazy load components and select which columns you want.

Fixes: exe-1329

## Type of change

## How Has This Been Tested?

Unit tests.

### DIFF
--- a/src/pixelator/pixeldataset/precomputed_layouts.py
+++ b/src/pixelator/pixeldataset/precomputed_layouts.py
@@ -87,9 +87,16 @@ class PreComputedLayouts:
         """
         return self._partitioning
 
-    def to_df(self) -> pd.DataFrame:
-        """Return the layouts as a pandas DataFrame."""
-        return self.lazy.collect().to_pandas()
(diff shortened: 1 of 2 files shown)


src/pixelator/pixeldataset/precomputed_layouts.py
```diff
@@ -87,9 +87,16 @@ def partitioning(self) -> list[str]:
         """
         return self._partitioning
 
-    def to_df(self) -> pd.DataFrame:
-        """Return the layouts as a pandas DataFrame."""
-        return self.lazy.collect().to_pandas()
+    def to_df(self, columns: list[str] | None = None) -> pd.DataFrame:
+        """Return the layouts as a pandas DataFrame.
+
+        :param columns: the columns to return, if `None` all columns will be returned
```
ambarrio, line 93: I see how this works, but it is a bit counter-intiutive. Good that is documented!

ambarrio (approved):

PR #100, opened 2024-03-01
Fix a bug in `is_read_file`

fbdtemme:
read1 and read2 suffixes where checked in the whole filename instead of just the end to deal with illumina suffixes like in `SampleName_S1_L001_R1_001.fastq.gz`. This can causes issues when multiple patterns are recognised. We now check for a 3 digit suffix specificly and anchor all patterns to the end.

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes: EXE-1468

## Type of change

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

New tests in `tests/utils/test_utils.py`

### DIFF
--- a/src/pixelator/report/common/workdir.py
+++ b/src/pixelator/report/common/workdir.py
@@ -312,8 +312,7 @@ class PixelatorWorkdir:
         fn: typing.Callable[[], dict[str, list[Path]]],
         sample: Optional[str] = None,
         cache: bool = True,
-    ) -> list[Path]:
-        ...
+    ) -> list[Path]: ...
 
     @typing.overload
     def _cached_reports_implementation(
@@ -322,8 +321,7 @@ class PixelatorWorkdir:
         fn: typing.Callable[[], dict[str, Path]],
         sample: Optional[str] = None,
         cache: bool = True,
-    ) -> Path:
-        ...
+    ) -> Path: ...
 
     def _cached_reports_implementation(
         self,
@@ -354,8 +352,7 @@ class PixelatorWorkdir:
         sample: None = None,
         *,
         cache: bool = True,
-    ) -> List[Path]:
-        ...
+    ) -> List[Path]: ...
 
     @typing.overload
     def single_cell_report(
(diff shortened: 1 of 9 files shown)


src/pixelator/report/common/workdir.py
johandahlberg, line 1: Maybe we should do a ruff-format on everything to not get these changes bit by bit in different PRs?

johandahlberg (approved):
Nice. And congrats on PR 100! 😄

src/pixelator/utils.py
```diff
@@ -431,19 +431,26 @@ def is_read_file(read: str, read_type: Literal["r1"] | Literal["r2"]) -> bool:
...
+        raise AssertionError("Invalid read type: expected 'r1' or 'r2'")
+
     if not (read.endswith("fq.gz") or read.endswith("fastq.gz")):
         raise ValueError("Invalid file extension: expected .fq.gz or .fastq.gz")
 
     matches = []
     read_stem = Path(read.removesuffix(get_extension(read, 2)).rstrip(".")).name
     if read_type == "r1":
-        matches = re.findall(R"(.[Rr]1|(_[Rr]?1))", read_stem)
+        matches = re.findall(R"(.[Rr]1$)|(_[Rr]?1$)|(_[Rr]?1_[0-9]{3}$)", read_stem)
```
ambarrio, line 445: Regarding this, it is for the Illumina numbered errors, right?

ambarrio (approved):
Just one checkin comment. Thanks for this!

src/pixelator/utils.py
fbdtemme: Yes this is to support things like: `RDRS182520_S1_L001_R1_001.fastq.gz`. Now that we always match from the end of the string we need to add this 3 digit segment.

src/pixelator/report/common/workdir.py
fbdtemme: I did a ruff on everything already using latest ruff to sync the local git hooks with the CI checks.

PR #102, opened 2024-03-01
Fixing the colocalization plot styles

ptajvar:
## Description

The functions plot_colocalization_heatmap and plot_colocalization_diff_heatmap are affected slightly to conform with the existing tutorial in the software site.

Changes:
- Bug fix: In differential colocalization we now report target - reference, before it was the other way resulting in inverted values. 
- Functional: Instead of getting the difference between median of colocalization values, we now use the median of all colocalization value-pair differences between the reference and the target.
- Cosmetic: elements in both are separated by white lines and in plot_colocalization_diff_heatmap the range is adjusted so that the center is 0.

Fixes EXE-1463

## Type of change

## How Has This Been Tested?

Re-running the unit tests and comparing the output plots.

## PR checklist:

- [x] New and existing unit tests pass locally with my changes


src/pixelator/analysis/colocalization/__init__.py
```diff
@@ -314,13 +332,31 @@ def get_differential_colocalization(
...
+                marker_data.name,
+                marker_data,
+                reference=reference,
+                target=target,
+                contrast_column=contrast_column,
+                value_column=value_column,
+            )
+        )
+    )
+    differntial_colocalization = colocalization_test_results.rename(
```
johandahlberg, line 348: typo: `differntial` -> `differential`

johandahlberg (approved):
Found a typo, other than that it LGTM.

src/pixelator/analysis/colocalization/__init__.py
ptajvar: Good catch, thank you!

PR #104, opened 2024-03-05
Feature/exe 1484 fix pmds performance

johandahlberg:
## Description

PMDS was computing shortest paths of all nodes vs all notes instead of only between the pivots. This

Fixes: exe-1484

## Type of change

## How Has This Been Tested?

Tried by manually plotting a few components, and they look ok.


ludvigla (approved):
Everything checks out. I tried the code on a simulated graph and it returns a sphere-like layout as expected:
simulated_graph_pmds_layout

PR #105, opened 2024-03-06
plot_3d_graph set to default color when no marker is given.

ptajvar:
Small PR just setting the default color when no marker is given in plot_3d_graph.


johandahlberg (approved):
I think it would be good to add a test to cover this case where `marker=None`. Other than that, this looks good.

PR #107, opened 2024-03-07
Update docker containers build workflow

fbdtemme:
## Description

This PR adds various containers related bits.

- Add some standard labels following the opencontainers annotations spec to the docker images and update the GH actions workflow to insert auto generated labels from the repo information.
- The base image is updated to fedora:39
- Build the package first in the container to make sure versions are reported properly

Fixes: EXE-1482

### DIFF
--- a/containers/base.Dockerfile
+++ b/containers/base.Dockerfile
@@ -108,14 +108,20 @@ RUN pip3.11 install --prefix=/runtime -r requirements.txt && rm requirements.txt
 # ------------------------------------------
 FROM runtime-base as build-pixelator
 
+ARG VERSION_OVERRIDE
+
 WORKDIR /pixelator
 COPY . /pixelator
 COPY .git /pixelator/.git
 
-RUN poetry config virtualenvs.create false && \
-    poetry dynamic-versioning && \
-    poetry build -f sdist && \
-    cp -r /pixelator/dist/ /dist/ && \
+RUN poetry config virtualenvs.create false
+RUN if [ ! -z "${VERSION_OVERRIDE}" ]; then \
+        echo "Overriding version to ${VERSION_OVERRIDE}"; \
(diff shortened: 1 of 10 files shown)


containers/base.Dockerfile
```diff
@@ -68,6 +70,7 @@ FROM builder-base as poetry-deps-install-amd64
 
 WORKDIR /pixelator
 COPY poetry.lock pyproject.toml ./
+COPY .git .git
```
johandahlberg, line 73: Didn't we talk about setting the via an environment variable instead? I haven't investigated it in practice, but I guess one thing we could do is use something like the below from the poetry dynamic versioning docs (https://pypi.org/project/poetry-dynamic-versioning/):

env `POETRY_DYNAMIC_VERSIONING_OVERRIDE="pixelator={{VERSION BASED ON GITHUB CI VARS}}"`.

On the other hand since the git folder is not included in the final container image anyway maybe this approach is easier?

johandahlberg (approved):
I just had some questions, but nothing that blocks merging this. Great job!

containers/base.Dockerfile
fbdtemme: I looked into using POETRY_DYNAMIC_VERSIONING_OVERRIDE like we do in eg. conda recipes, but I found this to be much more straigthforward. Especially when using this as a base image for pixelator-internal where we do want the detailed versioning.

I will add the option to still override the version with POETRY_DYNAMIC_VERSIONING_OVERRIDE using a build arg.

containers/base.Dockerfile
```diff
@@ -20,6 +20,8 @@ RUN python3.11 -m ensurepip
 RUN pip3.11 install --upgrade pip
 RUN pip3.11 install pipx
 RUN pipx install poetry
+RUN poetry self add poetry-plugin-export
+RUN poetry self add "poetry-dynamic-versioning[plugin]"
```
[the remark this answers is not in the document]
johandahlberg: Aha. I get it.

PR #109, opened 2024-03-08
Fixes in utils and improved multiprocess logging

johandahlberg:
## Description

Better logging in multiprocessing scenarios. 

Also allows some of our sanity check functions to take a single `PathType` to make them a bit more user friendly.

Fixes: EXE-1496

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Tested as part of the work I've been doing on the `layout` stage (a PR is incoming shortly).

PR #110, opened 2024-03-08
Add layout stage

johandahlberg:
## Description

This adds a new `layout` stage to pixelator, and includes a number of fixes relating to how layouts are generated to make sure we can deal with more scenarios than I originally had thought of.

This depends on #109 

Fixes: EXE-1332

## Type of change

## How Has This Been Tested?

Unit tests and running the command manually (I will do the later more before we finish up all of this).

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,7 +10,7 @@
         "editor.formatOnSave": true,
         "editor.defaultFormatter": "charliermarsh.ruff",
         "editor.codeActionsOnSave": {
-            "source.organizeImports": "explicit"
+            "source.organizeImports": "never"
         }
     },
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Provide typed interfaces for all per-stage report files using pydantic.
 * Centralize pixelator intermediate file lookup and access.
 * Add a `precomputed_layouts` property to `pixeldataset` to allow for loading precomputed layouts.
+* Add `pixelator single-cell layout` stage to pixelator, which allows users to compute layouts
+  for a PXL file that can then be used to visualize the graph in 2D or 3D downstream.
 
 ### Changed
 
--- a/src/pixelator/cli/layout.py
+++ b/src/pixelator/cli/layout.py
@@ -78,8 +78,8 @@ def layout(
     sanity_check_inputs(pxl_file, allowed_extensions="pxl")
 
     # create output folder if it does not exist
-    output_path = Path(output)
-    layout_output_dir = create_output_stage_dir(output_path, "layout")
+    workdir = PixelatorWorkdir(output)
+    layout_output_dir = workdir.stage_dir("layout")
 
     logger.info(f"Computing layout(s) for file {pxl_file}")
 
--- a/src/pixelator/cli/layout.py
+++ b/src/pixelator/cli/layout.py
@@ -42,7 +42,7 @@ from pixelator.utils import (
     required=False,
     is_flag=True,
     default=False,
-    help="Skip adding marker counts to the layout. This makes things a little less resource intensive. Default: False.",
+    help="Skip adding marker counts to the layout. Default: False.",
 )
 @click.option(
     "--layout-algorithm",
--- a/src/pixelator/cli/layout.py
+++ b/src/pixelator/cli/layout.py
@@ -1,0 +1,107 @@
+"""
+Console script for pixelator (layout)
+
+Copyright © 2024 Pixelgen Technologies AB.
+"""
+
+from pathlib import Path
+from typing import get_args
+
+import click
+
+from pixelator import read
+from pixelator.cli.common import logger, output_option
+from pixelator.graph.backends.protocol import SupportedLayoutAlgorithm
+from pixelator.pixeldataset.precomputed_layouts import (
+    generate_precomputed_layouts_for_components,
+)
+from pixelator.report.models.layout import LayoutSampleReport
+from pixelator.utils import (
+    create_output_stage_dir,
+    get_sample_name,
+    log_step_start,
+    sanity_check_inputs,
+    timer,
+    write_parameters_file,
+)
+
+
+@click.command(
+    "layout",
+    short_help="compute graph layouts that can be used to visualize components",
+    options_metavar="<options>",
+)
+@click.argument(
+    "pxl_file",
+    nargs=1,
+    required=True,
+    type=click.Path(exists=True),
+    metavar="PXL",
+)
+@click.option(
+    "--add-node-marker-counts",
+    required=False,
+    is_flag=True,
+    default=True,
+    help="Add marker counts to the layout. Default: True",
+)
+@click.option(
+    "--layout",
+    required=False,
+    multiple=True,
+    default=["pmds_3d"],
+    help="Select a layout algorithm to use. This can be specified multiple times to compute multiple layouts. Default: pmds_3d",
+    type=click.Choice(get_args(SupportedLayoutAlgorithm)),
+)
+@output_option
+@click.pass_context
+@timer
+def layout(
+    ctx,
+    pxl_file,
+    add_node_marker_counts,
+    layout,
+    output,
+):
+    """
+    Compute graph layouts that can be used to visualize components
+    """
+    log_step_start(
+        "layout",
+        input_files=pxl_file,
+        add_node_marker_counts=add_node_marker_counts,
+        layout=layout,
+        output=output,
+    )
+
+    # some basic sanity check on the input files
+    sanity_check_inputs(pxl_file, allowed_extensions="pxl")
+
+    # create output folder if it does not exist
+    output_path = Path(output)
+    layout_output_dir = create_output_stage_dir(output_path, "layout")
+
+    logger.info(f"Computing layout(s) for file {pxl_file}")
+
+    clean_name = get_sample_name(pxl_file)
+    write_parameters_file(
+        ctx,
+        layout_output_dir / f"{clean_name}.meta.json",
+        command_path="pixelator single-cell layout",
+    )
+
+    pxl_dataset = read(pxl_file)
+    pxl_dataset.precomputed_layouts = generate_precomputed_layouts_for_components(
+        pxl_dataset,
+        add_node_marker_counts=add_node_marker_counts,
+        layout_algorithms=layout,
+    )
+    pxl_dataset.save(
+        layout_output_dir / f"{clean_name}.layout.dataset.pxl", force_overwrite=True
+    )
+
+    metrics_file = layout_output_dir / f"{clean_name}.report.json"
+    report = LayoutSampleReport(
+        sample_id=clean_name,
+    )
+    report.write_json_file(metrics_file, indent=4)
--- a/src/pixelator/cli/layout.py
+++ b/src/pixelator/cli/layout.py
@@ -4,7 +4,6 @@ Console script for pixelator (layout)
 Copyright © 2024 Pixelgen Technologies AB.
 """
 
-from pathlib import Path
 from typing import get_args
 
 import click
@@ -15,9 +14,9 @@ from pixelator.graph.backends.protocol import SupportedLayoutAlgorithm
 from pixelator.pixeldataset.precomputed_layouts import (
     generate_precomputed_layouts_for_components,
 )
+from pixelator.report.common import PixelatorWorkdir
 from pixelator.report.models.layout import LayoutSampleReport
 from pixelator.utils import (
-    create_output_stage_dir,
     get_sample_name,
     log_step_start,
     sanity_check_inputs,
@@ -39,14 +38,14 @@ from pixelator.utils import (
     metavar="PXL",
 )
 @click.option(
-    "--add-node-marker-counts",
+    "--no-node-marker-counts",
     required=False,
     is_flag=True,
-    default=True,
-    help="Add marker counts to the layout. Default: True",
+    default=False,
+    help="Skip adding marker counts to the layout. This makes things a little less resource intensive. Default: False.",
 )
 @click.option(
-    "--layout",
+    "--layout-algorithm",
     required=False,
     multiple=True,
     default=["pmds_3d"],
@@ -59,8 +58,8 @@ from pixelator.utils import (
 def layout(
     ctx,
     pxl_file,
-    add_node_marker_counts,
-    layout,
+    no_node_marker_counts,
+    layout_algorithm,
     output,
 ):
     """
@@ -69,8 +68,8 @@ def layout(
     log_step_start(
         "layout",
         input_files=pxl_file,
-        add_node_marker_counts=add_node_marker_counts,
-        layout=layout,
+        no_node_marker_counts=no_node_marker_counts,
+        layout_algorithm=layout_algorithm,
         output=output,
     )
 
@@ -93,8 +92,8 @@ def layout(
     pxl_dataset = read(pxl_file)
     pxl_dataset.precomputed_layouts = generate_precomputed_layouts_for_components(
         pxl_dataset,
-        add_node_marker_counts=add_node_marker_counts,
-        layout_algorithms=layout,
+        add_node_marker_counts=not no_node_marker_counts,
+        layout_algorithms=layout_algorithm,
     )
     pxl_dataset.save(
         layout_output_dir / f"{clean_name}.layout.dataset.pxl", force_overwrite=True
--- a/src/pixelator/graph/backends/implementations/_networkx.py
+++ b/src/pixelator/graph/backends/implementations/_networkx.py
@@ -380,17 +380,15 @@ class NetworkXGraphBackend(GraphBackend):
         counts to use that can be used for plotting.
 
         The layout options are:
+          - pmds
+          - pmds_3d
           - fruchterman_reingold
           - fruchterman_reingold_3d
           - kamada_kawai
           - kamada_kawai_3d
-          - pmds
-          - pmds_3d
 
-        The `fruchterman_reingold` options are in general faster, but less
-        accurate than the `kamada_kawai` ones.
-
-        TODO Update these docs!
+        The `pmds` options are much faster than the `fruchterman_reingold` and
+        the `kamada_kawai`.
 
         :param layout_algorithm: the layout algorithm to use to generate the coordinates
         :param only_keep_a_pixels: If true, only keep the a-pixels
--- a/src/pixelator/graph/backends/implementations/_networkx.py
+++ b/src/pixelator/graph/backends/implementations/_networkx.py
@@ -20,6 +20,7 @@ from typing import (
     TYPE_CHECKING,
     Tuple,
     Union,
+    get_args,
 )
 
 import networkx as nx
@@ -37,6 +38,7 @@ from pixelator.graph.backends.protocol import (
     Vertex,
     VertexClustering,
     VertexSequence,
+    SupportedLayoutAlgorithm,
 )
 
 if TYPE_CHECKING:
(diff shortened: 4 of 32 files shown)


src/pixelator/cli/layout.py
```diff
@@ -0,0 +1,107 @@
...
+        layout=layout,
+        output=output,
+    )
+
+    # some basic sanity check on the input files
+    sanity_check_inputs(pxl_file, allowed_extensions="pxl")
+
+    # create output folder if it does not exist
+    output_path = Path(output)
+    layout_output_dir = create_output_stage_dir(output_path, "layout")
```
fbdtemme, line 82: I think we should phase this utility function out and replace it with PixelatorWorkdir.stage_dir

```suggestion
    workdir = PixelatorWorkdir(output)
    layout_output_dir = workdir.stage_dir("layout")
```

src/pixelator/cli/layout.py
```diff
@@ -0,0 +1,107 @@
...
+    options_metavar="<options>",
+)
+@click.argument(
+    "pxl_file",
+    nargs=1,
+    required=True,
+    type=click.Path(exists=True),
+    metavar="PXL",
+)
+@click.option(
```
fbdtemme, line 41: I think boolean flags should always be False by default. Passing true or false explicitly to a flag always feels weird to me.
So I would rename this to --no-node-marker-counts or something like that.

src/pixelator/cli/layout.py
```diff
@@ -0,0 +1,107 @@
...
+    "--add-node-marker-counts",
+    required=False,
+    is_flag=True,
+    default=True,
+    help="Add marker counts to the layout. Default: True",
+)
+@click.option(
+    "--layout",
```
fbdtemme, line 49: I think that this would be better renamed to `--method` or `--algorithm` since the subcommand is already named `layout`.

fbdtemme (commented):
This looks great.
I just have some naming nitpicks for the CLI interface.

fbdtemme (approved):

src/pixelator/pixeldataset/precomputed_layouts.py
```diff
@@ -33,8 +33,169 @@ class PreComputedLayoutsEmpty(PixelatorBaseException):
     """Raised when trying to access an empty PreComputedLayouts instance."""
 
 
+class RequestedLayoutsHaveMultipleDimensionalities(PixelatorBaseException):
+    """Raised when trying to materialize PreComputedLayouts with multiple dimensionalities."""
+
+
+class _DataProvider(Protocol):
```
ambarrio, line 40: __init__ method here not needed compare to `_EmptyDataProvider`?

src/pixelator/pixeldataset/precomputed_layouts.py
```diff
@@ -33,8 +33,169 @@ class PreComputedLayoutsEmpty(PixelatorBaseException):
...
+    ) -> list[pl.LazyFrame]:
+        raise PreComputedLayoutsEmpty("No layouts available")
+
+
+class _SingleFrameDataProvider(_DataProvider):
+    def __init__(self, lazy_frame: pl.LazyFrame) -> None:
+        self._lazy_frame = lazy_frame
+
+    def is_empty(self) -> bool:
+        return self.lazy().select(pl.col("component")).first().collect().is_empty()
```
ambarrio, line 88: As we are using in all the other ones?
```suggestion
        return self._lazy_frame.select(pl.col("component")).first().collect().is_empty()
```

src/pixelator/pixeldataset/precomputed_layouts.py
```diff
@@ -33,8 +33,169 @@ class PreComputedLayoutsEmpty(PixelatorBaseException):
...
+        if columns:
+            return self._lazy_frame.select(columns).collect().to_pandas()
+        return self._lazy_frame.collect().to_pandas()
+
+    def lazy(self):
+        return self._lazy_frame
+
+    def unique_components(self):
+        return set(
+            self.lazy()
```
ambarrio, line 100:
```suggestion
            self._lazy_frame
```

src/pixelator/pixeldataset/precomputed_layouts.py
```diff
@@ -33,8 +33,169 @@ class PreComputedLayoutsEmpty(PixelatorBaseException):
...
+
+    def to_df(self, columns: list[str] | None = None) -> pd.DataFrame: ...
+
+    def lazy(self): ...
+
+    def unique_components(self): ...
+
+    def filter(
+        self,
+        component_ids: str | set[str] | None = None,
```
ambarrio, line 51: I wonder: is this a set[str] or a list[str]?

src/pixelator/pixeldataset/precomputed_layouts.py
```diff
@@ -33,8 +33,169 @@ class PreComputedLayoutsEmpty(PixelatorBaseException):
...
+class _SingleFrameDataProvider(_DataProvider):
+    def __init__(self, lazy_frame: pl.LazyFrame) -> None:
+        self._lazy_frame = lazy_frame
+
+    def is_empty(self) -> bool:
+        return self.lazy().select(pl.col("component")).first().collect().is_empty()
+
+    def to_df(self, columns: list[str] | None = None) -> pd.DataFrame:
+        if columns:
+            return self._lazy_frame.select(columns).collect().to_pandas()
```
ambarrio, line 92: It would be actually good to do the opposite:
```suggestion
            return self.lazy().select(columns).collect().to_pandas()
```

src/pixelator/pixeldataset/precomputed_layouts.py
```diff
@@ -33,8 +33,169 @@ class PreComputedLayoutsEmpty(PixelatorBaseException):
...
+        return self._lazy_frame.collect().to_pandas()
```
ambarrio, line 93:
```suggestion
        return self.lazy().collect().to_pandas()
```

src/pixelator/pixeldataset/precomputed_layouts.py
```diff
@@ -253,81 +387,94 @@ def data():
...
+def _compute_layouts(
     edgelist: pl.DataFrame,
-    add_node_marker_counts,
-    layout_algorithm,
-    all_markers,
+    add_node_marker_counts: bool,
+    layout_algorithms: list[SupportedLayoutAlgorithm] | SupportedLayoutAlgorithm,
+    all_markers: list[str],
 ):
+    if isinstance(layout_algorithms, str):
```
ambarrio, line 403: As we have typed this `layout_algorithms`, is this really a `str`or a `SupportedLayoutAlgorithm`?

src/pixelator/pixeldataset/precomputed_layouts.py
```diff
@@ -253,81 +387,94 @@ def data():
...
+                            cache=False,
+                        )
+                        .pipe(
+                            _zero_fill_missing_markers
+                            # If we don't want to add the marker counts
+                            # we just pass this through an identity function
+                            if add_node_marker_counts
+                            else lambda df, all_markers: df,
+                            all_markers=all_markers,
+                        )
```
ambarrio, line 430: Interesting expression. That'd be cool to understand. Or simplify if people are going to be same interpretation at I do.

src/pixelator/pixeldataset/precomputed_layouts.py
```diff
@@ -253,81 +387,94 @@ def data():
...
 
-    result = pl.concat(data())
+    result = pl.concat(data(), how="diagonal")
     return result
 
 
-def _fn(d):
-    (edgelist, add_node_marker_counts, layout_algorithm, all_markers) = d
-    return _get_layout(edgelist, add_node_marker_counts, layout_algorithm, all_markers)
+def _wrap_get_layouts(d):
```
ambarrio, line 445: Wouldn't work *args here?

src/pixelator/pixeldataset/precomputed_layouts.py
```diff
@@ -337,12 +484,21 @@ def _data(
...
+    layout_algorithms: SupportedLayoutAlgorithm
+    | list[SupportedLayoutAlgorithm] = "pmds_3d",
 ) -> PreComputedLayouts:
+    """Generate precomputed layouts for the components in the PixelDataset.
+
+    :param pixel_dataset: the PixelDataset to generate the layouts for
+    :param components: the components to generate the layouts for, if `None` all components will be used
+    :param add_node_marker_counts: whether to add the marker counts to the layout. If you don't need them
+                                   you can set this to false and speed up the computations.
+    :param layout_algorithm: the layout algorithm to use
```
ambarrio, line 500:
```suggestion
    :param layout_algorithms: the layout algorithm(s) to use
```

ambarrio (changes requested):

.vscode/settings.json
```diff
@@ -10,7 +10,7 @@
         "editor.formatOnSave": true,
         "editor.defaultFormatter": "charliermarsh.ruff",
         "editor.codeActionsOnSave": {
-            "source.organizeImports": "explicit"
+            "source.organizeImports": "never"
```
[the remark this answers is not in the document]
johandahlberg: I think @fbdtemme doesn't use vscode, so these settings shouldn't matter to him. I do sort them, but only manually, because when the setting above is not "never", there is some bug in the vscode plugin that makes it remove code. Which is rather annoying...

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Provide typed interfaces for all per-stage report files using pydantic.
 * Centralize pixelator intermediate file lookup and access.
 * Add a `precomputed_layouts` property to `pixeldataset` to allow for loading precomputed layouts.
+* Add `pixelator single-cell layout` stage to pixelator, which allows users to compute layouts
+  for a PXL file that can then be used to visualize the graph in 2D or 3D downstream.
```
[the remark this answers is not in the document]
johandahlberg: Right now we do `pmds_3d` per default from the `layout` command.

tests/pixeldataset/test_precomputed_layouts.py
```diff
@@ -38,13 +39,46 @@ def layout_df_fixture() -> pd.DataFrame:
...
 
-@pytest.fixture(name="precomputed_layouts")
-def precomputed_layouts_fixture(layout_df) -> pd.DataFrame:
-    yield PreComputedLayouts(pl.DataFrame(layout_df).lazy())
+# We are using this to make sure we cover both cases of the PreComputedLayouts
+# one there we have a DataFrame and one where we have a generator of data frames
+@pytest.fixture(name="precomputed_layouts", params=["layout_df", "layout_df_generator"])
+def precomputed_layouts_fixture(request) -> pd.DataFrame:
+    if request.param == "layout_df":
+        return PreComputedLayouts(layout_df())
```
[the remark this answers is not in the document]
johandahlberg: The typing is wrong. I'll fix it.

tests/pixeldataset/test_precomputed_layouts.py
```diff
@@ -313,3 +357,165 @@ def test_aggregate_precomputed_layouts_no_layouts_in_data(self):
...
+
+        The overhead of running in separate processes is not worth it here.
+        """
+
+        class MockPoolExecutor:
+            def __init__(self):
+                pass
+
+            def imap(self, func, *args, **kwargs):
+                yield from map(func, *args)
```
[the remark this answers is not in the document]
johandahlberg: Yes, exactly. It is to match the signature of `imap`.

tests/pixeldataset/test_precomputed_layouts.py
```diff
@@ -313,3 +357,165 @@ def test_aggregate_precomputed_layouts_no_layouts_in_data(self):
...
+        ) as mock_pool_executor:
+            mock_pool_executor.return_value = MockPoolExecutor()
+            yield mock_pool_executor
+
+    @pytest.fixture(name="pixel_dataset")
+    def pixel_dataset_fixture(self, setup_basic_pixel_dataset):
+        (dataset, *_) = setup_basic_pixel_dataset
+        yield dataset
+
+    def test_generate_precomputed_layouts_for_components_with_all_components(
```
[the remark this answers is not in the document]
johandahlberg: It follows the convention of using `test_<name of function under test>_<info about test>`. Hence it becomes a bit repetitive here since components come up in both the function name and the test info.

tests/report/assets/reports_only/layout/pbmcs_unstimulated.report.json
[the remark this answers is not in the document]
johandahlberg: No. Right now I don't know what would be useful to add to the report, so I suggest keeping it like this for now. And then we can add metrics as we realize what we want to put in there.

src/pixelator/cli/layout.py
```diff
@@ -0,0 +1,106 @@
...
+    is_flag=True,
+    default=False,
+    help="Skip adding marker counts to the layout. This makes things a little less resource intensive. Default: False.",
+)
+@click.option(
+    "--layout-algorithm",
+    required=False,
+    multiple=True,
+    default=["pmds_3d"],
+    help="Select a layout algorithm to use. This can be specified multiple times to compute multiple layouts. Default: pmds_3d",
```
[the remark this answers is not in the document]
johandahlberg: No, if you want to pass it multiple times you add multiple times like so: `--layout-algorithm one_method --layout-algorithm another_method`

src/pixelator/graph/backends/implementations/_networkx.py
```diff
@@ -386,15 +380,15 @@ def layout_coordinates(
...
           - fruchterman_reingold
           - fruchterman_reingold_3d
           - kamada_kawai
           - kamada_kawai_3d
-          - pmds
-          - pmds_3d
 
-        The `fruchterman_reingold` options are in general faster, but less
-        accurate than the `kamada_kawai` ones.
+        The `pmds` options are much faster than the `fruchterman_reingold` and
```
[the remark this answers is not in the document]
johandahlberg: 10-100x faster. I can add that.

src/pixelator/graph/backends/protocol.py
```diff
@@ -148,6 +158,8 @@ def layout_coordinates(
           - pmds_3d
 
 
+        TODO Rewrite the docs here to indicate that we use pmds
```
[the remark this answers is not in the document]
johandahlberg: Yes. Well spotted!

src/pixelator/pixeldataset/backends.py
```diff
@@ -284,6 +284,7 @@ def precomputed_layouts(self) -> PreComputedLayouts | None:
         """Get the precomputed layouts."""
         # If it is None it means it is uninitialized, and we should
         # attempt to read it lazily
+
         if self._precomputed_layouts is None:
```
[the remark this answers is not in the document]
johandahlberg: I'll fix it.

src/pixelator/pixeldataset/datastores.py
```diff
@@ -461,13 +462,9 @@ def write_precomputed_layouts(
...
 
-        # Note that the raw_iterator may contain multiple partitions
-        # that will depend on how the objects were initiated. But, this
-        # will be handled by the write_dataframe method.
-        for layouts_to_write in layouts.raw_iterator:
-            logger.debug("Writing layout...")
+        for layouts_to_write in layouts.component_iterator():
             self.write_dataframe(
-                layouts_to_write.collect().to_pandas(),
+                layouts_to_write,
```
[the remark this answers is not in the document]
johandahlberg: Exactly.

src/pixelator/pixeldataset/datastores.py
```diff
@@ -461,13 +462,9 @@ def write_precomputed_layouts(
 
         self._check_if_writeable(self.LAYOUTS_KEY)
 
-        # Note that the raw_iterator may contain multiple partitions
-        # that will depend on how the objects were initiated. But, this
-        # will be handled by the write_dataframe method.
-        for layouts_to_write in layouts.raw_iterator:
-            logger.debug("Writing layout...")
+        for layouts_to_write in layouts.component_iterator():
```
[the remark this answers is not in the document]
johandahlberg: The `debug` statement here has been removed. I removed when I was cleaning things up here since it became rather noisy.

src/pixelator/pixeldataset/datastores.py
```diff
@@ -554,6 +551,17 @@ def read_dataframe_lazy(self, key: str) -> Optional[pl.LazyFrame]:
             "as a parquet based pxl file."
         )
 
+    def write_precomputed_layouts(
+        self,
+        layouts: Optional[PreComputedLayouts],
+    ) -> None:
+        """Write pre-computed layouts to the data store (NB: Not implemented!)."""
+        raise NotImplementedError(
+            "You are trying to write precomputed layouts to a csv-based pxl file. "
```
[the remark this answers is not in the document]
johandahlberg: Yeah, but if so let's do it in a separate PR to keep these changes a bit cleaner.

src/pixelator/pixeldataset/precomputed_layouts.py
johandahlberg: Nope.

src/pixelator/cli/layout.py
```diff
@@ -0,0 +1,106 @@
...
+    required=True,
+    type=click.Path(exists=True),
+    metavar="PXL",
+)
+@click.option(
+    "--no-node-marker-counts",
+    required=False,
+    is_flag=True,
+    default=False,
+    help="Skip adding marker counts to the layout. This makes things a little less resource intensive. Default: False.",
```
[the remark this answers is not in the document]
johandahlberg: I'll remove the sentence about the resources.

src/pixelator/pixeldataset/precomputed_layouts.py
johandahlberg: Good catch.

src/pixelator/pixeldataset/precomputed_layouts.py
johandahlberg: The way I understand literals they are only a type checking mechanism, at runtime they are just their concert types, i.e. `str` in this case. You can read more about Literal types here: https://typing.readthedocs.io/en/latest/spec/literal.html#literal-types

src/pixelator/pixeldataset/precomputed_layouts.py
johandahlberg: This is basically equivalent with the psudo-code below:

```
def f(df: pd.DataFrame, add_node_marker_counts: bool):
   if add_node_marker_counts:
       return _zero_fill_missing_markers(df)
   return df
```
But written with https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.pipe.html I think it mad sense here since in general we are just chaining functions on each other here.

src/pixelator/pixeldataset/precomputed_layouts.py
johandahlberg: It would, but I wanted to make it clear which inputs `_compute_layouts` expects.

src/pixelator/pixeldataset/datastores.py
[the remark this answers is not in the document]
ambarrio: Ok, I was trying to check we didn't track the looping in case it would die there.

tests/pixeldataset/test_precomputed_layouts.py
```diff
@@ -313,3 +357,165 @@ def test_aggregate_precomputed_layouts_no_layouts_in_data(self):
...
+            "x",
+            "y",
+            "z",
+            "x_norm",
+            "y_norm",
+            "z_norm",
+            "graph_projection",
+            "layout",
+            "component",
+        } | set(pixel_dataset.adata.var.index)
```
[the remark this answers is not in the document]
ambarrio: Understandably! 😄

ambarrio (approved):

PR #112, opened 2024-03-14
Enforce import sorting

johandahlberg:
## Description

This PR adds enforcing of import sort order according to the `ruff` `isort` rules. It also auto fixes this across all files. This should have no functional impact on the code.

Fixes: exe-1512


fbdtemme (approved):
:+1:  I was worried about circular imports but I saw that you already disabled it where necessary with the `# isort: off ` comment.

PR #113, opened 2024-03-15
Feature/exe 1514 cache graph marker count

ptajvar:
Makes node_marker_count a cached_property of graph to avoid reconstructing the dataframe with every new call. The property will be a dataframe that is indexed with node ids rather than names. The main reason for this is to enable getting marker counts separately from neighborhood marker calculations. This in turn enables getting neighborhood markers after node and marker filtering and also speed up parameter sweeps for analysis.

prepare_from_edgelist is now removed from colocalization prepare.

Fixes: EXE-1514

## Type of change

## How Has This Been Tested?

Several existing unit tests including layout calculations use this property.

## PR checklist:


johandahlberg (approved):
Excellent!

PR #117, opened 2024-03-22
Draft refactoring of analysis

johandahlberg:
This is not actually meant to be merged... I just wanted to put up a draft of a potential refactoring of the analysis module that I have been thinking about.

The idea here is to encapsulate different per component analysis functions as classes, and allow developers to implement new ones by implementing the `PerComponentAnalysis`. Then you would only have to deal with the actual analysis code, not with any of the surrounding things like creating the graph, setting up multiprocessing etc.

Other than making it easier to add new methods my initial benchmarks on this demonstrates that this is roughly ~6x faster than the current implementation. This is archived by to main changes. Firstly, the `Graph` object is only created once, and then passed to the different analysis steps. Secondly, it puts these methods on different processes.

The main drawback of this approach as I see it is that `Graph` instance creation is not parallelized, and it of course also has the overhead of having to pickle/unpickel the data as it is sent back and forth to the subprocesses.

I'd be happy to hear your thoughts about this approach and if you have any suggestions for improvements.

### DIFF
--- a/src/pixelator/analysis/refactored_analysis.py
+++ b/src/pixelator/analysis/refactored_analysis.py
@@ -1,0 +1,220 @@
+import logging
+from collections import defaultdict
+from functools import partial
+from queue import Queue
+from threading import get_ident
+from typing import Callable, Iterable, Protocol
+
+import pandas as pd
+
+from pixelator.analysis.colocalization import colocalization_from_component_graph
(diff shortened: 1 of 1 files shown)


johandahlberg:
Actual implementation lives in #127

PR #125, opened 2024-04-11
fix removal of wrong read 1/2 id when multiple matches are present

fbdtemme:
## Description

Fix a bug where the wrong read 1/2 suffix is stripped away causing the pipeline to file due to mismatches sample names 
between r1 and r2 read pairs.

Fixes: EXE-1633


johandahlberg (approved):
LGTM

PR #136, opened 2024-05-02
Added local_g

ludvigla:
## Description

Adding `compute_transition_probabilities` and `local_g` functions. `compute_transition_probabilities` takes an adjacency matrix as input and returns a matrix with transition probabilities between node pairs for a k-step random walk. `local_g` takes an adjacency matrix and a marker count matrix as input and returns a DataFrame with local G scores per marker and node. By default, `local_g` uses transition probabilities for a k-step random walk (`compute_transition_probabilities`) to construct a weighting scheme for the metric computation. 

The main expected use case of local G scores is for visualizing spatial trends of marker expression on 2D/3D layouts.

`local_g` is provided as a separate function within the new `node_metrics` module and can also be called as a method on Graph objects.

Fixes: EXE-1677

## Type of change

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce it when relevant.
- Added two tests for `compute_transition_probabilities` with the parameter `remove_self_loops` set to `True` or `False` (test_node_metric.py)
- Added four tests for `local_g` (test_node_metric.py)
   1. Default parameter settings
   2. `use_weights=False`
   3. `normalize_counts=False`
   4. `method='gstari'`
- Added one test for `Graph` method `local_g`  with default settings. (test_graph.py)

## PR checklist:

### DIFF
--- a/src/pixelator/graph/graph.py
+++ b/src/pixelator/graph/graph.py
@@ -299,7 +299,7 @@ class Graph:
     ) -> pd.DataFrame:
         """Compute the local G metric for each node in the graph.
 
-                The local G metric is a measure of the local clustering of a node in the graph.
+        The local G metric is a measure of the local clustering of a node in the graph.
 
         :param k: The number of steps in the k-step random walk. Default is 1.
         :param use_weights: Whether to use weights in the computation. When turned off, all
--- a/src/pixelator/graph/graph.py
+++ b/src/pixelator/graph/graph.py
@@ -7,11 +7,11 @@ from __future__ import annotations
 
 import logging
 from functools import cached_property, lru_cache
-from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
+from typing import Any, Dict, Iterable, List, Literal, Optional, Tuple, Union
 
 import pandas as pd
 import polars as pl
-from scipy.sparse import csr_matrix
+from scipy.sparse import csr_array, csr_matrix
 
 from pixelator.graph.backends.implementations import (
     graph_backend,
@@ -22,6 +22,7 @@ from pixelator.graph.backends.protocol import (
     SupportedLayoutAlgorithm,
     VertexClustering,
 )
+from pixelator.graph.node_metrics import local_g
 
 logger = logging.getLogger(__name__)
 
@@ -210,7 +211,7 @@ class Graph:
         :param only_keep_a_pixels: If true, only keep the a-pixels
         :param get_node_marker_matrix: Add a matrix of marker counts to each
                                        node if True.
-        :param cache: set to `True` in order to cache one call of this this method.
+        :param cache: set to `True` in k to cache one call of this this method.
                       It will make subsequent calls to the layout method
                       with the same settings much faster, at the cost of additional
                       memory usage. This can speed things up a lot when plotting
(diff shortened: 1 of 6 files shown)


src/pixelator/graph/graph.py
```diff
@@ -210,7 +211,7 @@ def layout_coordinates(
         :param only_keep_a_pixels: If true, only keep the a-pixels
         :param get_node_marker_matrix: Add a matrix of marker counts to each
                                        node if True.
-        :param cache: set to `True` in order to cache one call of this this method.
+        :param cache: set to `True` in k to cache one call of this this method.
```
johandahlberg, line 214:
```suggestion
        :param cache: set to `True` in order to cache one call of this this method.
```

src/pixelator/graph/graph.py
```diff
@@ -287,3 +288,25 @@ def add_names_to_vertexes(self, vs_names: List[str]) -> None:
...
+        self,
+        k: int = 1,
+        use_weights: bool = True,
+        normalize_counts: bool = True,
+        W: csr_array | None = None,
+        method: Literal["gi", "gistar"] = "gi",
+    ) -> pd.DataFrame:
+        """Compute the local G metric for each node in the graph.
+
+        The local G metric is a measure of the local clustering of a node in the graph.
```
johandahlberg, line 302:
```suggestion
        The local G metric is a measure of the local clustering of a node in the graph.
        
        :param k: The number of steps in the k-step random walk. Default is 1.
        :param use_weights: Whether to use weights in the computation. When turned off, all
        edge weights will be equal to 1. Default is True.
        :param normalize_counts: Whether to normalize counts to proportions. Default is True.
        :param W: A sparse matrix of custom edge weights. This will override the automated
        computation of edge weights. `W` must have the same dimensions as A. Note that weights can
        be defined for any pair of nodes, not only the pairs represented by edges in `A`. Default is None.
        :param method: The method to use for computing local G. Must be one of 'gi' or 'gistar'.
        'gi' is the original local G metric, which does not consider self-loops, meaning that the
        local marker expression for a node is computed by aggregating the weighted expression of
        its neighbors. 'gistar' is a simplified version of local G that does consider self-loops.
        In other words, the local marker expression of a node also includes the weighted marker
        expression of the node itself. Default is 'gi'.
        :return: A DataFrame of local G-scores for each node and marker.
```

tests/graph/test_graph.py
```diff
@@ -311,7 +311,7 @@ def test_connected_components_benchmark(benchmark, enable_backend, edgelist):
 def test_get_adjacency_sparse(enable_backend, pentagram_graph):
     # This is a little bit involved. Since different network backends might
     # use different internal indexing schemes, they are not guaranteed to generate
-    # the same order of nodes in the adjacency matrix
+    # the same k of nodes in the adjacency matrix
```
johandahlberg, line 314:
```suggestion
    # the same order of nodes in the adjacency matrix
```

johandahlberg (approved):
Some doc update suggestions. Otherwise superb work!

src/pixelator/graph/graph.py
ludvigla: Thanks! Completely missed this :-)

PR #138, opened 2024-05-14
Adding the rate_diff_transformation function.

ptajvar:
rate_diff_transformation is an alternative transformation that has been shown to improve the raw pearson values for colocalization analysis especially in terms of removing the correlation between colocalization values and marker abundance. It is now being added as an option but will not be set as the default transformation yet for colocalization analysis (the default is currently log1p). 

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Added a unit-test for colocalization calculation with rate-diff transformation to verify run without errors. The colocalization values with rate-diff transformation are also checked for different datasets and are shown to result in high agreement between raw pearson values and pearson_z scores and expected biological signals such as HLA-ABC with B2M and CD11a and CD18 appear consistently in top marker pairs. 
There will be further inspection and comparison before setting this transformation as default.

## PR checklist:

- [X] This comment contains a description of changes (with reason).

### DIFF
--- a/src/pixelator/statistics.py
+++ b/src/pixelator/statistics.py
@@ -131,6 +131,22 @@ def log1p_transformation(df: pd.DataFrame) -> pd.DataFrame:
     return log1p_df
 
 
+def rate_diff_transformation(df: pd.DataFrame) -> pd.DataFrame:
+    """Transform antibody counts based on deviation from an expected baseline distribution.
+
+    In this function we refer to baseline distribution as fixed ratio of different antibody types in each node. For example, if in total 10% of antibodies are HLA-ABC, in a node with 120 antibodies we expect to see 12 HLA-ABC counts. If we actually see 8 counts in this node, the rate_diff_transformation for HLA-ABC in this node will be -4.
+    :param df: the dataframe of raw antibody counts (antibodies as columns)
+    :returns: a dataframe with the counts difference from expected values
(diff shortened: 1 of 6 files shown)


src/pixelator/statistics.py
```diff
@@ -131,6 +131,22 @@ def log1p_transformation(df: pd.DataFrame) -> pd.DataFrame:
     return log1p_df
 
 
+def rate_diff_transformation(df: pd.DataFrame) -> pd.DataFrame:
+    """Transform antibody counts based on deviation from an expected baseline distribution.
+
+    In this function we refer to baseline distribution as fixed ratio of different antibody types in each node. For example, if in total 10% of antibodies are HLA-ABC, in a node with 120 antibodies we expect to see 12 HLA-ABC counts. If we actually see 8 counts in this node, the rate_diff_transformation for HLA-ABC in this node will be -4.
```
johandahlberg, line 137: We should break this line up into lines that are roughly 80 characters long. Unfortunately ruff in unable to autoformat this.

johandahlberg (changes requested):
This looks really nice and clean! I had a small comment on the formatting of the doc sting. The other thing I think would be good to do is to add a simple test for `rate_diff_tranformation`. Lets fix that and then I think we are good to go.

johandahlberg (approved):
Super!

PR #147, opened 2024-05-31
Feature/exe 1769 dsb implementation

ptajvar:
## Description

Added dsb_normalization as part of the new normalization package in the analysis stage. dsb normalization has been shown to effectively compensate for noise effects in expression data.

Fixes: #(issue number)

## Type of change

## How Has This Been Tested?

It has been tested against a similar function that is already implemented in R and the results are almost identical (A slight variation is expected given that the gaussian mixture model stage is indeterministic). 
A corresponding test function is implemented to compare against future changes to the function.
## PR checklist:


src/pixelator/analysis/normalization/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+from typing import List, Union
+
+import numpy as np
+import pandas as pd
+from sklearn.decomposition import PCA
+from sklearn.mixture import GaussianMixture
+
+
+def _limma(pheno, exprs, rcond=1e-8):
+    """Linear regression to remove confounding factors from expression data."""
```
johandahlberg, line 15: Maybe we should call it abundance data here?

src/pixelator/analysis/normalization/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+            marker_data = dataframe.loc[i, :].to_frame()
+            gmm = gmm.fit(marker_data)
+            baseline[i] = np.min(gmm.means_)
+            scores[i] = np.abs(gmm.means_[1] - gmm.means_[0]) / np.sum(gmm.covariances_)
+    elif axis == 1:
+        for i in dataframe.columns:
+            marker_data = dataframe.loc[:, i].to_frame()
+            gmm = gmm.fit(marker_data)
+            baseline[i] = np.min(gmm.means_)
+            scores[i] = np.abs(gmm.means_[1] - gmm.means_[0]) / np.sum(gmm.covariances_)
```
johandahlberg, line 38: I think perhaps we could simplify this to something like:

```
if axis not in {0, 1}:
    raise ValueError(f"Axis was {axis}. Must be 0 or 1")
pick_axis = lambda df: df.loc[i, :] if axis == 0 else lamba: df.loc[:, i]
for i in dataframe.index:
    marker_data = pick_axis(dataframe).to_frame()
    gmm = gmm.fit(marker_data)
    baseline[i] = np.min(gmm.means_)
    scores[i] = np.abs(gmm.means_[1] - gmm.means_[0]) / np.sum(gmm.covariances_)


```

src/pixelator/analysis/normalization/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+    return baseline, scores
+
+
+def dsb_normalize(
+    raw_expression: pd.DataFrame, isotype_controls: Union[List, None] = None
+):
+    """empty-droplet-free method as implemented in the dsb package.
```
johandahlberg, line 45: I think we should add a reference to the original package here.

tests/analysis/normalization/test_normalization.py
```diff
@@ -0,0 +1,26 @@
...
+from pathlib import Path
+
+import pandas as pd
+from pandas.testing import assert_frame_equal
+from pixelator.analysis.normalization import dsb_normalize
+
+DATA_ROOT = Path(__file__).parents[2] / "data"
+
+
+def test_dsb_normalize():
```
johandahlberg, line 16: Perhaps we should also add a test where `isotype_controls=None` to make sure we cover that code path.

src/pixelator/analysis/normalization/__init__.py
johandahlberg: I haven't actually test this.

johandahlberg (approved):
Excellent! I have focuses on the python aspect here. I think @maxkarlsson is more qualified to comment on the actual implementation of the method.

Very nice and clear code. I had some minor suggestions.

src/pixelator/analysis/normalization/__init__.py
maxkarlsson: Mulè, M.P., Martins, A.J. & Tsang, J.S. Normalizing and denoising protein expression data from droplet-based single cell profiling. Nat Commun 13, 2099 (2022).
https://doi.org/10.1038/s41467-022-29356-8

src/pixelator/analysis/normalization/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+    :param raw_abundance: the raw abundance count data.
+    :param isotype_controls: list of isotype controls.
+    :return: normalized abundance data.
+    """
+    log_abundance = np.log1p(raw_abundance)
+    marker_background, _ = _get_background_abundance(log_abundance, axis=1)
+    log_abundance = log_abundance - marker_background
+    component_background, _ = _get_background_abundance(log_abundance, axis=0)
+
+    if isotype_controls is not None:
```
maxkarlsson, line 57: I think we should force the user to input isotypes, since dsb relies on those.

src/pixelator/analysis/normalization/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+        control_signals = log_abundance.loc[:, isotype_controls]
+        control_signals["component_background"] = component_background
+        pheno = PCA(n_components=1).fit_transform(control_signals)
```
maxkarlsson, line 60: Are variables in `control_signals` scaled to unit variance before PCA?

maxkarlsson (commented):
Looks great! I have some comments and a control question for the PCA step. I have a little trouble understanding whether the linear regression is equivalent, mostly because I am not used to the syntax. I would suggest that we generate data for both versions of the normalization and verify that the result is close to exactly the same. Great and quick work!

I also made some commits with terminology changes. Please hava a look and see if you agree with the changes.

src/pixelator/analysis/normalization/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+
+from typing import List, Union
+
+import numpy as np
+import pandas as pd
+from sklearn.decomposition import PCA
+from sklearn.mixture import GaussianMixture
+
+
+def _limma(pheno, exprs, rcond=1e-8):
```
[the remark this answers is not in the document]
maxkarlsson: I changed this now in the new commit

src/pixelator/analysis/normalization/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+    scores = pd.Series(index=dataframe.index if axis == 0 else dataframe.columns)
+    gmm = GaussianMixture(n_components=2, max_iter=1000, random_state=0)
+    if axis not in {0, 1}:
+        raise ValueError(f"Axis was {axis}. Must be 0 or 1")
+    ax_iter = dataframe.index if axis == 0 else dataframe.columns
+    for i in ax_iter:
+        current_axis = dataframe.loc[i, :] if axis == 0 else dataframe.loc[:, i]
+        gmm = gmm.fit(current_axis.to_frame())
+        background[i] = np.min(gmm.means_)
+        scores[i] = np.abs(gmm.means_[1] - gmm.means_[0]) / np.sum(gmm.covariances_)
```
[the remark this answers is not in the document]
ptajvar: Yes! right now the score is not used anywhere as we do the baseline subtraction for all markers, but could later be used as a check for existence of negative population.

maxkarlsson (approved):
Looks great!

PR #148, opened 2024-06-04
Handle when child process is killed by OOM killer

johandahlberg:
## Description

This PR adds checking to `collapse` for when child processes are killed (most likely by the OMM killer), and if that is the case has the main process exit with `137` as it's error code. This allows e.g. nextflow to detect this as a process that can be retried with additional memory. 

Fixes: exe-1777

## Type of change

## How Has This Been Tested?

I have tried this manually by making the child process build an infinite list until they crash with OOM, and made sure that the main process exits gracefully with 137 as exit code.

## PR checklist:


src/pixelator/cli/collapse.py
```diff
@@ -25,6 +26,38 @@
...
+                    "The child process in the process pool returned a non-zero exit code: %s.",
+                    exit_code,
+                )
+                # If we have an out of memory exception, make sure we exit with that.
+                if abs(exit_code) == 9:
+                    logger.error(
+                        "One of the child processes was killed (exit code: 9). Usually this is caused "
+                        "by a child process using to much memory. We will return an exit code of 137 "
+                        "to indicate that the process was killed by the out of memory killer."
+                    )
```
fbdtemme, line 53: I think this is currently slightly inaccurate since anyone can send kill signals.
Maybe something like this?

```suggestion
                        "One of the child processes was killed (exit code: 9). "
                        "Usually this is caused by the out-of-memory killer terminating the process. "
                        "The parent process will return an exit code of 137 to indicate that it terminated because of a kill signal in the child process."
                    )
```

fbdtemme (approved):

PR #150, opened 2024-06-26
Feature/exe 1476 precom layouts for plots

ptajvar:
If plot_2d_graph or plot_3d_graph are called without specified layout_algorithm (or layout_algorithm=None), the precomputed_layout will be used if it exists. Otherwise pmds_3d will be used as the default layout.

Fixes: EXE-1476

## Type of change

## How Has This Been Tested?

test_plot_2d_graph_precomputed and test_plot_3d_graph_precomputed are added to test the new conditions. The old tests pass as well. 

## PR checklist:


src/pixelator/plot/__init__.py
```diff
@@ -545,9 +561,9 @@ def plot_3d_graph(
     pxl_data: PixelDataset,
     component: str,
     marker: Union[list, None] = None,
-    layout_algorithm: Literal[
-        "fruchterman_reingold_3d", "kamada_kawai_3d", "pmds_3d"
-    ] = "fruchterman_reingold_3d",
+    layout_algorithm: Union[
+        Literal["fruchterman_reingold_3d", "kamada_kawai_3d", "pmds_3d"], None
+    ] = None,
```
fbdtemme, line 566: I would just factor this out into a LayoutAlgorithms alias and then use `LayoutAlgorithms | None` here.
(Since we dropped python <3.10 support we can use the new union syntax).

The same goes for the input to `plot_2d_graph`.

src/pixelator/plot/__init__.py
```diff
@@ -392,8 +406,10 @@ def _get_color_values(mark, filtered_coordinates, log_scale, vmax):
 def plot_2d_graph(
     pxl_data: PixelDataset,
     component: Union["str", list],
```
fbdtemme, line 408: This type hint does not seem to match with the code

```suggestion
    component: Iterable[str]
```

fbdtemme (approved):
Just two small things but otherwise approved.

src/pixelator/plot/__init__.py
ptajvar: Should it be Union[str, Iterable[str]]? Because for the case where the user provides only one component we have:
if isinstance(component, str):
        component = [component]

PR #151, opened 2024-06-27
Implement low-count marker removal in pixelator colocalizaiton

ptajvar:
Exclude low-count markers in a component from colocalization calculation.

Fixes: EXE-1826
## Type of change

## How Has This Been Tested?

The existing tests pass. A new test is added to remove markers with low counts from a component and check how many marker-pairs are included in colocalization output.

## PR checklist:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -138,6 +138,16 @@ def random_graph_edgelist_fixture():
     return edgelist
 
 
+@pytest.fixture(name="full_random_graph_edgelist", scope="module")
+def full_random_graph_edgelist_fixture():
+    """Create an edgelist based on a random graph."""
+    g = create_random_graph(n_nodes=500, prob=0.005)
+    edgelist = create_simple_edge_list_from_graph(g, random_markers=True)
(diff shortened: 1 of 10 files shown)


tests/conftest.py
```diff
@@ -138,6 +138,16 @@ def random_graph_edgelist_fixture():
     return edgelist
 
 
+@pytest.fixture(name="full_random_graph_edgelist", scope="module")
+def full_random_graph_edgelist_fixture():
```
johandahlberg, line 142: This looks very similar to `random_graph_edgelist_fixture`. Why do we need this?

johandahlberg (approved):
LGTM! I had a question, but it's nothing blocking merging this.

tests/conftest.py
ptajvar: random_graph_edgelist_fixture randomizes the edges but not the markers (the first half are marker A and second half are marker B). full_random_graph_edgelist also randomizes markers from A to G and we need multiple markers to be able to check the effect of removing low-count markers on colocalization scores.

johandahlberg: Aha. You are right. Thanks.

PR #152, opened 2024-07-01
Add panel aliases, add a v2 placeholder panel

fbdtemme:
## Description

Add a new panel file metadata field: `aliases`.
This is a list of alternate names that can be used to query the `Config` object for panels.
The `pixelator single-cell --list-panels` option will now also list aliases.

Rename current panels to: `human-sc-immunology-spatial-proteomics-1`
Add a new placeholder panel: `human-sc-immunology-spatial-proteomics-2`

The `human-sc-immunology-spatial-proteomics-1` panels now use an alias to `human-sc-immunology-spatial-proteomics`
to make sure that all old samplesheets continue working as expected.

Fixes: EXE-1803

## Type of change

## How Has This Been Tested?

Updated config tests


src/pixelator/config/config_class.py
```diff
@@ -83,11 +107,40 @@ def get_assay(self, assay_name: str) -> Optional[Assay]:
...
+
+        :param include_aliases: Include panel aliases in the list
+        :returns: A list of panel names
+        """
+        out = sorted(list(self.panels.keys()))
+
+        if not include_aliases:
+            return out
+
+        out += sorted(list(self.panel_aliases.keys()))
```
johandahlberg, line 121: I thought perhaps it could be nice if names and aliases are listed together. Something like this?

```
pixelator single-cell --list-panels
human-sc-immunology-spatial-proteomics-1 (aliases: human-sc-immunology-spatial-proteomics, apa, monkey)
human-sc-immunology-spatial-proteomics-2
```

Not the cleanest implementation, but here is a suggestion:

```suggestion
    def list_panel_names(self, include_aliases: bool = False) -> List[str]:
        """Return a list of all panel names.

        :param include_aliases: Include panel aliases in the list
        :returns: A list of panel names
        """

        plain_out = sorted(list(self.panels.keys()))

        if not include_aliases:
            return plain_out

        def get_names_to_aliases():
            names_to_aliases = defaultdict(list)
            for aliases, name in self.panel_aliases.items():
                names_to_aliases[name].append(aliases)
            return names_to_aliases

        panel_names_to_aliases = get_names_to_aliases()

        out_with_aliases = []
        for name in plain_out:
            aliases = panel_names_to_aliases.get(name) or []
            if not aliases:
                out_with_aliases.append(name)
                continue
            aliases_str = ", ".join(aliases) if aliases else ""
            out_with_aliases.append(f"{name} (aliases: {aliases_str})")

        return out_with_aliases
```

johandahlberg (approved):
Nice. I had some minor suggestions that you can accept or reject at you see fit.

src/pixelator/config/config_class.py
fbdtemme: 100% agree, but this is currently mainly used in nf-core/pixelator to validate the possible options passed in the `panel` field. 

I was thinking of making the "nice output" version the default and adding something like `--raw` and use that in nf-core/pixelator but wanted to do that in a separate PR

johandahlberg: Aha! I see. Sounds reasonable.

ambarrio (commented):
LGTM. Great job!

PR #159, opened 2024-07-08
fix: bug in compute_transition_probabilities

ludvigla:
## Description

The transition probabilities are invalid at k > 1 in `compute_transition_probabilities` which can influence the computation of local G scores. This PR contains a fix for this bug along with some additional minor fixes:

### Changes

- correctly row-normalize `W_out` matrix after matrix power computation
- handle situation where `use_weights=False` to match R implementation
- set `normalize_counts=False` as default to match R implementation

Fixes: exe-1867

## Type of change

## PR checklist:


ambarrio (commented):
LGTM. Only comment is how did you discover this? Can you in this case include maybe a test to collect this knowledge that needs to always pass?

ludvigla:
> LGTM. Only comment is how did you discover this? Can you in this case include maybe a test to collect this knowledge that needs to always pass?

I discovered it when creating the local G tutorial for Python because the p-values didn't match with R. These things are quite difficult to detect which is why I think it's so important that more people play around with the method in their analysis. 

I will add a test!

johandahlberg (approved):
LGTM

PR #160, opened 2024-07-09
Split pixelator.plot into modules

ptajvar:
There are several plot functions being added so we are breaking them into functional modules. So far, spatial_analysis_plots and layout_plots are the introduced modules. This does not affect the user interface as the functions can still be imported directly from pixelator.plot. 

The only breaking change in this PR is that edge_rank_plot is renamed to molecule_rank_plot.

Fixes: EXE-1869

## Type of change

## How Has This Been Tested?

Existing tests pass. No new functionality is added.

## PR checklist:


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 * The default transformation for the calculation of the colocalization score is now `rate-diff` instead of `log1p`.
+* Rename `edge_rank_plot` function to `molecule_rank_plot`.
```
ludvigla, line 22: Would it make sense to keep `edge_rank_plot` too for backwards compatibility and remove it in a future release?

ptajvar: It should fail for any pixel file that is generated after 0.16.2 anyway because edge_rank_plot looks for 'edges' in adata.obs and that has been replaced 'molecules' since then. But we could keep it for older pixel files, maybe with future warning @johandahlberg?

johandahlberg: Yeah, I think it might be good to keep the old one around as well for now (with a deprecation notice as suggested). As this is probably a plot that people actually use.

We could also make it a bit more robust by falling back to `edges` if `molecules` is not available.

johandahlberg (approved):
Small suggestion, but you can decide how you want to deal with it. I'll approve this.

CHANGELOG.md
ludvigla: That sounds like a reasonable solution to me!

ludvigla (approved):
Looks good to me!

CHANGELOG.md
ptajvar: Now added the backward compatibility with the old function name and use of "edges" column instead of "molecules" in the new function.

PR #168, opened 2024-07-12
Allow wider version range for umi tools

johandahlberg:
## Description

Allow a wider range of versions for umi tools to allow us to build conda packages.


ambarrio (approved):
Great addition!

PR #177, opened 2024-08-01
Fix local development container builds running with PYTHONNOUSERSITE on AWS

fbdtemme:
## Description

After updating to a new Amazon Linux container the local development containers file to run on AWS Batch due to the
aws-cli trying to load a pixelator specific package instead of its own version.

The aws-cli is mounted inside the container by nextflow so we need to remove the use of `PYTHONPATH` since that will interfer with the aws-cli dependencies.

All packages installed to the custom `PYTHONPATH` location have been moved so the containers play nice again with nextflow
with `export PYTHONNOUSERSITE=1`.

## Type of change


ambarrio (approved):
LGTM

PR #180, opened 2024-08-20
Fix some typos and tweak development Dockerfile 

fbdtemme:
## Description

- Some cosmetic fixes in the assay file region names.
- Only install dev packages in the dev Container file. All main packages are already installed in the base layer


johandahlberg (approved):
LGTM

PR #183, opened 2024-08-22
add support for multiple targets in differential colocalization

ptajvar:
## Description

Adding support for multiple targets in both calculation and plotting of differential colocalizaiton. Now the function plot_colocalization_diff_volcano makes a figure with up to 5 subplots corresponding to different targets against the reference. plot_colocalization_diff_heatmap makes multiple figures instead of one with subplots as seaborn clustermap has its own subplots and is not (trivially) possible to merge with multiple other subplots. 

Fixes: EXE-1935

## Type of change

## How Has This Been Tested?

A new test is added for making volcano plots with multiple targets. For the clustermap plot, the old test is updated as it now generates a series of figures rather than a single figure containing all target plots.

## PR checklist:


src/pixelator/analysis/colocalization/__init__.py
```diff
@@ -341,20 +340,13 @@ def _colocalization_wilcoxon(
...
+    return pd.Series({"stat": stat, "p_value": p_value, "median_difference": estimate})
 
 
 def get_differential_colocalization(
     colocalization_data_frame: pd.DataFrame,
-    target: str,
     reference: str,
+    targets: str | list[str] | None = None,
     contrast_column: str = "sample",
     use_z_score: bool = True,
```
ludvigla, line 351: If we add a different score in the future, e.g. conditional probability or join-counts, would it make sense to enable setting this manually with a `value_column ` argument?

src/pixelator/plot/spatial_analysis_plots.py
```diff
@@ -101,102 +108,100 @@ def plot_colocalization_heatmap(
...
+    contrast_column.
     :param reference: The label for reference components in the contrast_column.
+    :param targets: label or list of labels for target components in the
+    contrast_column. When not specified, all labels in the contrast_column
+    except the reference label are used as targets.
     :param contrast_column: The column to use for the contrast. Defaults to "sample".
-    :param markers: The markers to include in the heatmap. Defaults to None. At most only one of n_top_marker_pairs or markers should be provided.
-    :param n_top_marker_pairs: The number of top marker pairs to include in the heatmap. Defaults to None. At most only one of n_top_marker_pairs or markers should be provided.
+    :param top_marker_log_p: When set to a value, only markers with at least
+    one higher log10 p-score are plotted.
```
ludvigla, line 136: Not sure if I understand this. Only markers with log10 p-value > `top_marker_log_p` are plotted?

ludvigla (approved):
Looks good! Added some minor comments

src/pixelator/analysis/colocalization/__init__.py
ptajvar: Good point, we can change the input to be value_column directly since we are already changing function inputs anyway.

src/pixelator/plot/spatial_analysis_plots.py
ptajvar: Yes, markers with differential colocalization with log10 p-value > top_marker_log_p with at least one other marker since the p-values are assigned to marker-pairs. I will rephrase it to be more explicit.

PR #184, opened 2024-08-22
Fix running tests with tox and rework testing workflow

fbdtemme:
## Description

Update testing gh workflows using tox.

- Streamline the use of poetry and the way the package is installed.
- Use the tox-gh plugin instead of a manual script to map workflow matrix versions to tox versions


johandahlberg (approved):
LGTM

ambarrio (approved):

PR #186, opened 2024-08-30
Updated probability distance calculations

ludvigla:
## Description

[Edges in multigraphs](https://networkx.org/documentation/stable/reference/classes/generated/networkx.MultiGraph.edges.html) have an additional key that makes the weighted pMDS algorithm fail. Since we have used simple graphs previously, this was not handled by `_prob_edge_weights` which expected a tuple of two in each edge.

This PR contains a fix to enable running `_prob_edge_weights` regardless of whether the graph is simple or multigraph.

Moreover, the method has been updated to be more performant on larger graphs.

Fixes: exe-1997

## Type of change

## How Has This Been Tested?

Added a new test for `pmds_layout` with "prob_dist" weights on a multigraph version of `pentagram_graph`.

## PR checklist:

### DIFF
--- a/src/pixelator/graph/backends/implementations/_networkx.py
+++ b/src/pixelator/graph/backends/implementations/_networkx.py
@@ -775,7 +775,10 @@ def pmds_layout(
     Options are:
 
     * an np.array with non-negative values (same number of elements as edges in g)
-    * "prob_dist" to use -log(P)^3, where P is the probability of a random walker to traverse the end nodes of and edge (i->j) and back again (j->i) in 5 steps.
+    * "prob_dist" to weight each edge (i, j) by -log(P)^3, where P is the probability
+        of a random walker to go from i to j in 5 steps and then back again (j->i)
+        in 5 steps. For this computation, self-loops are added to the graph to ensure
+        that all transitions are possible.
     * None to use unweighted shortest path lengths
     :param seed: Set seed for reproducibility
     :return: A dataframe with layout coordinates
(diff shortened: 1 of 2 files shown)


ptajvar (approved):
LGTM!

src/pixelator/graph/backends/implementations/_networkx.py
```diff
@@ -904,18 +936,28 @@ def _prob_edge_weights(
...
+    nodes = np.array(list(g.nodes))
 
-    return np.array(edge_probs)[0]
+    # Get end node IDs for the edges
+    node_from = edges[:, 0]
+    node_to = edges[:, 1]
+
+    # Find the correct positions to extract from the probability matrix
+    index_dict = {val: idx for idx, val in enumerate(nodes)}
+    vectorized_index = np.vectorize(lambda x: index_dict.get(x, -1))
```
ptajvar, line 953: elegant!

src/pixelator/graph/backends/implementations/_networkx.py
```diff
@@ -904,18 +936,28 @@ def _prob_edge_weights(
...
     P_step_bidirectional = P_step.multiply(P_step.T)
 
     # Extract the transition probabilities for edges in g
-    edges = list(g.edges)
-    edge_probs = P_step_bidirectional[
-        [list(g.nodes).index(u) for u, _ in edges],
-        [list(g.nodes).index(v) for _, v in edges],
-    ]
+    edges = np.array(list(g.edges))
+    nodes = np.array(list(g.nodes))
```
johandahlberg, line 945: I don't think there is a need to make the `edge`/`nodes` into lists here. In fact I think skipping it should be slightly more performant since it should avoid an extra allocation of the data.

src/pixelator/graph/backends/implementations/_networkx.py
johandahlberg: Agreed!

johandahlberg (approved):
Small comment, but with that small fix I think this is good to go. I am approving it now.

PR #193, opened 2024-10-15
Adding annotation of potential doublets to anndata obs

ptajvar:
During the graph step, we split some components that appear to be technical multiplets through community detection. However we set parameters in a relatively conservative way so to avoid breaking up cells into multiple components. In the annotation step we are adding a less conservative parametrization to mark "potential doublets".

What is added:
- is_potential_doublet as a column in adata.
- n_edges_to_split_doublet as a column in adata: how many edges (molecules) should be removed to split the two (or more) detected sub-communities in the potential doublet.
- fraction_potential_doublets to the json report of the annotation step. (rate of is_potential_doublet)
- n_edges_to_split_potential_doublets to the json report of the annotation step. (sum of n_edges_to_split_doublet)

Fixes: EXE-2025

## Type of change

## How Has This Been Tested?

The unit tests.

## PR checklist:


src/pixelator/annotate/__init__.py
```diff
@@ -376,6 +376,8 @@ class AnnotateAnndataStatistics(typing.TypedDict):
     min_size_threshold: Optional[int]
     max_size_threshold: Optional[int]
     doublet_size_threshold: Optional[int]
+    fraction_potential_doublets: float
+    n_edges_to_split_potential_doublets: int
```
ambarrio, line 380: I wonder why these are not `Optional` and e.g. the aggregates metrics are. Do you know @fbdtemme ? I guess if it is not found, will it be `None` or `0`?

ambarrio: I wonder if we should set `None` here to initialize and differ from the case where we find 0 doublets.

src/pixelator/annotate/__init__.py
```diff
@@ -433,4 +437,12 @@ def anndata_metrics(adata: AnnData) -> AnnotateAnndataStatistics:
     if "doublet_size_threshold" in adata.uns:
         metrics["doublet_size_threshold"] = adata.uns["doublet_size_threshold"]
 
+    if "is_potential_doublet" in adata.obs:
+        metrics["fraction_potential_doublets"] = adata.obs[
```
ambarrio, line 441: If there are no `is_potential_doublet` here, what will it happen when run mean?

src/pixelator/annotate/__init__.py
```diff
@@ -433,4 +437,12 @@ def anndata_metrics(adata: AnnData) -> AnnotateAnndataStatistics:
...
+            "is_potential_doublet"
+        ].mean()
+        metrics["n_edges_to_split_potential_doublets"] = adata.obs[
```
ambarrio, line 444: Same with sum

src/pixelator/pixeldataset/utils.py
```diff
@@ -217,6 +219,76 @@ def write_anndata(adata: AnnData, filename: PathType) -> None:
...
+        .count()
+        .reset_index()
+        .sort_values(["upia", "upib"])
+    )
+    edgelist_tuple = list(
+        map(tuple, np.array(component_edgelist[["upia", "upib", "count"]]))
+    )
+    component_communities_dict = leiden(
+        edgelist_tuple,
+        resolution=0.5 * LEIDEN_RESOLUTION,
```
ambarrio, line 236: Maybe we can include a comment here on the decision we took in the ticket about doing 1/2 of the resolution

ambarrio: Maybe we can include it within a single-line docstring on what this function is going to achieve.

ambarrio (approved):
LGTM. Some comments to take into account, but approving to not delay a post review merge once they are answered or tackled.

src/pixelator/pixeldataset/utils.py
```diff
@@ -217,6 +219,76 @@ def write_anndata(adata: AnnData, filename: PathType) -> None:
...
+
+
+def mark_potential_doublets(
+    edgelist: pd.DataFrame,
+) -> pd.Series:
+    """Mark whether a component is a potential doublet.
+
+    A component is a potential doublet if a) it has more than one community and
+    b) the second largest community is at least 20% of the size of the largest
+    community. (If the other communities are smaller they are assumed to be debries.)
```
johandahlberg, line 274:
```suggestion
    community.
```

I suggest removing the comment about debris here. Depending on the connectivity of the component I don't think we can be sure that the communities represent debris or not.

src/pixelator/pixeldataset/utils.py
```diff
@@ -217,6 +219,76 @@ def write_anndata(adata: AnnData, filename: PathType) -> None:
...
+            != component_edgelist["upib"].map(component_communities)
+        ).sum()
+        return True, edges_to_remove
+    else:
+        return False, 0
+
+
+def mark_potential_doublets(
+    edgelist: pd.DataFrame,
+) -> pd.Series:
```
johandahlberg, line 269:
```suggestion
) -> tuple[pd.Series, pd.Series]:
```
I'm a bit unsure here, but shouldn't this be the type annotation?

src/pixelator/annotate/__init__.py
ptajvar: There will be a key error, that's why we check for it first.

ambarrio: But when `adata.obs["is_potential_doublet"]` returns empty and we called `mean()` - what happens then?

ptajvar: We only run the mean() function if "is_potential_doublet" exists as a column in adata.obs. Otherwise that part is not run and "fraction_potential_doublets" remains the default value which is 0 right now but following the discussion we're going to change it to None.

PR #200, opened 2024-12-10
Feature/exe 2075 fail if hard and dynamic size limits are both set

johandahlberg:
## Description

Fail if both dynamic and hard size limits are set in `annotate` at the same time.

Fixes: exe-2075

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Tested by manually running the cli.

## PR checklist:

- [x] This comment contains a description of changes (with reason).
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] If a new tool or package is included, I have updated poetry.lock, and [cited it properly](../CITATIONS.md)
- [x] I have checked my code and documentation and corrected any misspellings
- [x] I have documented any significant changes to the code in [CHANGELOG.md](../CHANGELOG.md)

PR #201, opened 2024-12-10
Feature/exe 2065 set hard cutoff on min size

johandahlberg:
## Description

Set a hard cutoff of 300 edges on the minimum component size when using `--dynamic-filter min` or `--dynamic-filter both`  in `annotate`. 

Depends on merging #200.

Fixes: exe-2065

## Type of change

## How Has This Been Tested?

Unit tests and by manually running pixelator.

## PR checklist:

- [x] If a new tool or package is included, I have updated poetry.lock, and [cited it properly](../CITATIONS.md)


tests/annotate/test_cell_calling.py
```diff
@@ -40,7 +41,7 @@ def test_find_component_min_size_limits_only_noise():
     test_data = random_state.poisson(100, 500)
 
     min_bound = find_component_size_limits(np.absolute(test_data), direction="lower")
-    assert min_bound < 100
+    assert min_bound >= MINIMUM_N_EDGES_CELL_SIZE
```
ambarrio, line 44: Is this going to work now? Do we have `test_data` with cells over 300 edges. I ask because the sign before was the opposite actually.

ambarrio (approved):
LGTM. I approve with just a question that can be answered.

tests/annotate/test_cell_calling.py
johandahlberg: I revered it since now we will never find a minimum value smaller than the threshold, even when we only have noise - which is what this test tries to get to.

PR #216, opened 2025-04-22
Moving dsb normalization to common

ptajvar:
dsb_normalize, similar to clr_transformation is an abundance related method that is common between MPX and PNA. We are therefore moving it to the corresponding module.

Fixes: PNA-785

## Type of change

## How Has This Been Tested?

There are no functional changes introduced. The corresponding unit test is moved together with the main function.

## PR checklist:


johandahlberg (approved):
I think we should note this change in the changelog, since it will change the imports on the MPX side. Other than that I think this looks good to go, so I will approve it now.

PR #232, opened 2025-05-21
Passing None through _memory_validator in demux

ptajvar:
We currently get an error when None is passed to _memory_validator. It is however expected to output None in that case.

## Type of change

## How Has This Been Tested?

Unit tests.

## PR checklist:

### DIFF
--- a/src/pixelator/common/utils/units.py
+++ b/src/pixelator/common/utils/units.py
@@ -8,12 +8,15 @@ import re
 units = {"K": 10**3, "M": 10**6, "G": 10**9}
 
 
-def parse_size(s: str) -> int | float:
+def parse_size(s: str | None) -> int | float | None:
     """Parse a string as a number with optional unit suffix [K, M, G].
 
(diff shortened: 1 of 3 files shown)


fbdtemme (commented):
Is something breaking when not passing `--memory`? The intention is that when passing nothing/None duckdb will use its default limits (80% of available memory) as a soft limit before starting to spill-over to tmp disk. I think the default auto detected limit here makes more sense than an hardcoded default.

ptajvar:
> Is something breaking when not passing `--memory`? The intention is that when passing nothing/None duckdb will use its default limits (80% of available memory) as a soft limit before starting to spill-over to tmp disk. I think the default auto detected limit here makes more sense than an hardcoded default.

When not passing --memory we get TypeError: expected string or bytes-like object, got 'NoneType'. It seems to come from _memory_validator when returning int(parse_size(value)).

fbdtemme (approved):

PR #233, opened 2025-05-21
Move PNA->MPX dependencies to the common module

ptajvar:
PNA functions should be independent from MPX ones. We are now moving existing dependencies to the "common" module.

Fixes: PNA-721

## Type of change

## How Has This Been Tested?

Unit tests and ensure that pixelator.mpx does not appear in the pna module or the common module.
TODO: full pipeline run.

## PR checklist:


src/pixelator/mpx/report/models/__init__.py
```diff
@@ -26,5 +25,4 @@
     "MoleculesDataflowReport",
     "CommandInfo",
     "CommandOption",
-    "SummaryStatistics",
```
johandahlberg, line 29: Shouldn't this also export `SummaryStatistics`? Since we seem to export all the other classes here?

src/pixelator/pna/cli/layout.py
```diff
@@ -119,6 +119,8 @@ def layout(
     metrics_file = layout_output_dir / f"{clean_name}.report.json"
     report = LayoutSampleReport(
         sample_id=clean_name,
+        product_id="single-cell-pna",
+        report_type="layout",
     )
```
johandahlberg, line 124: I don't understand why this would need to change, since it is only swapping the import?

johandahlberg (approved):
Great work! Some minor questions and comments, but nothing blocking merging.

src/pixelator/pna/cli/layout.py
ptajvar: The MPX base SampleReport report doesn't have product_id and report_type but PNA SampleReport has them. Before we were using the MPX one to generate the layout sample report.

src/pixelator/mpx/report/models/__init__.py
ptajvar: SummaryStatistics is now a common model. We could import it here from common and export it as well if not having it here may break things.

johandahlberg: Super! Thanks.

src/pixelator/pna/cli/layout.py
johandahlberg: Ok. Missed that. Thanks for the clarification.

PR #238, opened 2025-06-04
Feature/pna 944 support single end reads

ptajvar:
Adding support for single-end reads.

Fixes: PNA-944

## Type of change

## How Has This Been Tested?

Added unit tests with single-end reads.

## PR checklist:

### DIFF
--- a/src/pixelator/pna/amplicon/build_amplicon.py
+++ b/src/pixelator/pna/amplicon/build_amplicon.py
@@ -527,8 +527,6 @@ class AmpliconBuilder(CombiningModifier, HasFilterStatistics, HasCustomStatistic
         :param region1_slice: the slice from the forward read
         :param region2_slice: the slice from the reverse read
         """
-        assert read1.qualities is not None
-        assert read2.qualities is not None
 
         if not region1_slice and not region2_slice:
             return None, None
@@ -626,11 +624,6 @@ class AmpliconBuilder(CombiningModifier, HasFilterStatistics, HasCustomStatistic
         # Since we only use this for Q30 statistics, this is not a big issue
         # For fully accurate quality scores we would need to have the full alignment
         # instead of just the matched region and this has a very high overhead.
-        if read1 is not None:
-            assert read1.qualities is not None
-        if read2 is not None:
-            assert read2.qualities is not None
-
         if not region1_slice and not region2_slice:
             return template_qual
 
--- a/src/pixelator/pna/amplicon/build_amplicon.py
+++ b/src/pixelator/pna/amplicon/build_amplicon.py
@@ -778,6 +778,8 @@ class AmpliconBuilder(CombiningModifier, HasFilterStatistics, HasCustomStatistic
 
 @dataclass
 class Amplicon:
+    """A dataclass to hold the regions of an amplicon."""
+
     pid1_umi1_region_seq: bytearray | Any | bytes | None
     pid1_umi1_region_qual: bytearray | Any | None
     pid2_umi2_region_seq: bytearray | Any | bytes | None
@@ -793,17 +795,31 @@ class PairedEndAmpliconBuilder(AmpliconBuilder):
 
     def handle(
         self,
-        read1: SequenceRecord | None,
-        read2: SequenceRecord | None,
+        read1: SequenceRecord,
+        read2: SequenceRecord,
         info1: ModificationInfo | None = None,
         info2: ModificationInfo | None = None,
     ) -> tuple[Amplicon, AmpliconBuilderFailureReason | None]:
-        if read1 is None and read2 is None:
-            return None
+        """Process paired-end sequencing reads to build an Amplicon object and determine failure reasons.
 
-        # Create a slice for each region in the amplicon (if it could be found)
-        r1_regions = self._scan_forward_read(read1) if read1 is not None else None
-        r2_regions = self._scan_reverse_read(read2) if read2 is not None else None
+        This method scans the provided forward and reverse reads for specific regions, generates consensus
+        sequences and quality scores for each region, and checks for errors. If consensus cannot be reached,
+        it returns an Amplicon object with all regions set to None and an appropriate failure reason.
+
+        Args:
+            read1 (SequenceRecord): The forward sequencing read.
+            read2 (SequenceRecord): The reverse sequencing read.
+            info1 (ModificationInfo | None, optional): Additional modification information for read1. Defaults to None.
+            info2 (ModificationInfo | None, optional): Additional modification information for read2. Defaults to None.
+
+        Returns:
+            tuple[Amplicon, AmpliconBuilderFailureReason | None]:
+                A tuple containing the constructed Amplicon object and a failure reason if applicable.
+                If both reads are None, returns None.
+
+        """
+        r1_regions = self._scan_forward_read(read1)
+        r2_regions = self._scan_reverse_read(read2)
         try:
             # Combine the info from forward and reverse reads, or use only the available read
             pid1_umi1_region_seq, pid1_umi1_region_qual = self._consensus_seq(
@@ -867,6 +883,8 @@ class PairedEndAmpliconBuilder(AmpliconBuilder):
 
 
 class SingleEndAmpliconBuilder(AmpliconBuilder):
+    """A wrapper for AmpliconBuilder that handles single-end reads."""
+
     def handle(
         self,
         read1: SequenceRecord | None,
@@ -874,6 +892,19 @@ class SingleEndAmpliconBuilder(AmpliconBuilder):
         info1: ModificationInfo | None = None,
         info2: ModificationInfo | None = None,
     ) -> tuple[Amplicon, AmpliconBuilderFailureReason | None]:
+        """Process a single read or a pair of reads to construct an amplicon.
+
+        Args:
+            read1 (SequenceRecord | None): The forward read, or None if not available.
+            read2 (SequenceRecord | None): The reverse read, or None if not available.
+            info1 (ModificationInfo | None, optional): Modification info for the forward read (ignored).
+            info2 (ModificationInfo | None, optional): Modification info for the reverse read (ignored).
+
+        Returns:
+            tuple[Amplicon, AmpliconBuilderFailureReason | None]:
+                A tuple containing the constructed Amplicon and an optional failure reason.
+
+        """
         is_read1 = read1 is not None
         read = read1 if is_read1 else read2
 
--- a/src/pixelator/pna/amplicon/build_amplicon.py
+++ b/src/pixelator/pna/amplicon/build_amplicon.py
@@ -560,10 +560,10 @@ class AmpliconBuilder(CombiningModifier, HasFilterStatistics, HasCustomStatistic
                 read2, region2_slice, is_reversed=True, as_bytearray=True
             )
 
-            assert len(s1) == len(s2)
-            assert len(q1) == len(q2)
+            assert len(s1) == len(s2)  # type: ignore
+            assert len(q1) == len(q2)  # type: ignore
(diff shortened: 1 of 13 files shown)


src/pixelator/pna/amplicon/build_amplicon.py
```diff
@@ -666,60 +708,118 @@ def _check_regions(
...
+            )
+            lbs2_region_qual = self._consensus_qual_lbs2(
+                read1,
+                read2,
+                region1_slice=regions.lbs2 if is_read1 else None,  # type: ignore
+                region2_slice=regions.lbs2 if not is_read1 else None,  # type: ignore
+            )
+            error = self._check_regions(
+                pid1_umi1_region_seq, uei_region_seq, pid2_umi2_region_seq
+            )
```
johandahlberg, line 768: Perhaps we can break this into a utility function? I think this function feels a bit long.

src/pixelator/pna/amplicon/build_amplicon.py
```diff
@@ -666,60 +708,118 @@ def _check_regions(
...
+                )
 
-        # Check for errors and increment the statistics
-        error = self._check_regions(
-            pid1_umi1_region_seq, uei_region_seq, pid2_umi2_region_seq
-        )
+                # Check for errors and increment the statistics
+                error = self._check_regions(
+                    pid1_umi1_region_seq, uei_region_seq, pid2_umi2_region_seq
+                )
```
johandahlberg, line 806: I'd suggest placing this into a utility function as well.

src/pixelator/pna/amplicon/report.py
```diff
@@ -272,3 +275,18 @@ def _collect_step(self, step):
                 self._custom_stats[name] += step.get_statistics()
             else:
                 self._custom_stats[name] = step.get_statistics()
+
+    def collect(
+        self,
+        n: int,
+        total_bp1: int,
+        total_bp2: Optional[int],
```
johandahlberg, line 283:
```suggestion
        total_bp2: int | None,
```

src/pixelator/pna/demux/report.py
```diff
@@ -62,6 +62,21 @@ def _collect_step(self, step):
             else:
                 self._custom_stats[name] = step.get_statistics()
 
+    def collect(
+        self,
+        n: int,
+        total_bp1: int,
+        total_bp2: Optional[int],
```
johandahlberg, line 69:
```suggestion
        total_bp2: int | None,
```

src/pixelator/pna/demux/report.py
```diff
@@ -3,7 +3,7 @@
 Copyright © 2024 Pixelgen Technologies AB.
 """
 
-from typing import Any, Dict
+from typing import Any, Dict, Optional
```
johandahlberg, line 6:
```suggestion
from typing import Any, Dict
```

src/pixelator/pna/amplicon/report.py
```diff
@@ -4,7 +4,7 @@
 """
 
 import typing
-from typing import Any, Dict
+from typing import Any, Dict, Optional
```
johandahlberg, line 7:
```suggestion
from typing import Any, Dict
```

src/pixelator/pna/read_processing/pipeline.py
```diff
@@ -52,10 +52,12 @@
 
 
 class AmpliconPipeline(Pipeline):
-    """Processing pipeline for combining paired-end reads into a single amplicon sequence.
+    """Processing pipeline for combining paired‐end reads now extended to also accept single‐end input (either R1 _or_ R2).
```
johandahlberg, line 55:
```suggestion
    """Processing pipeline for processing single-end or paired‐end reads into a single amplicon sequence.
```

src/pixelator/pna/read_processing/pipeline.py
```diff
@@ -64,27 +66,44 @@ class AmpliconPipeline(Pipeline):
...
-        pre_modifiers: Iterable[PairedEndModifier] | None = None,
-        pre_steps: Iterable[PairedEndStep] | None = None,
+        pre_modifiers: (
+            Iterable[
+                Union[
+                    PairedEndModifier,
+                    SingleEndModifier,
+                    tuple[SingleEndModifier | None, SingleEndModifier | None],
+                ]
+            ]
```
johandahlberg, line 81: I'm wondering if perhaps this type annotation can be simplified into `PairedEndModifier | SingleEndModifier | tuple[SingleEndModifier | None, SingleEndModifier | None]`? I think `Union` and `|` do the same thing, so mixing them makes it a bit more difficult t read do me.

src/pixelator/pna/read_processing/pipeline.py
```diff
@@ -64,27 +66,44 @@ class AmpliconPipeline(Pipeline):
...
         :param pre_modifiers: A list of modifiers that are applied to the reads before the combining step.
         :param pre_steps: A list of steps that are applied to the reads before the combining step.
         :param post_modifiers: A list of modifiers that are applied to the reads after the combining step.
         :param post_steps: A list of steps that are applied to the reads after the combining step.
         """
         self._combiner = combiner
-        self._pre_modifiers: list[PairedEndModifier] = []
-        self._pre_steps: list[PairedEndStep] = list(pre_steps) if pre_steps else []
+        self._pre_modifiers: list[
+            Union[
```
johandahlberg, line 100: Same question here as above.

src/pixelator/pna/read_processing/pipeline.py
```diff
@@ -150,6 +179,63 @@ def _add_modifier(self, modifier: PairedEndModifier, stage: ModifierStage) -> No
...
+        if reads is not None:
+            read1, read2 = reads
+
+        return read1, read2, info1, info2, n_bp1, n_bp2
+
+    def _pre_process_single(self, single_read):
+        """Pre‐process single reads."""
+        pre_modifiers_and_steps = self._pre_modifiers + self._pre_steps
+        name = single_read.name
+        is_r2 = ("R2" in name) or ("/2" in name) or ("_2" in name)
```
johandahlberg, line 205: `or ("/2" in name) or ("_2" in name)` I am not sure that these two conditions won't cause any issues. I don't  `/2` is a valid file name, and `_2` could cause issues if the used has a sample with a name like `sample_rep_2_...`. I suggest only looking for `R2` since I think that is the default from Illumina sequencers.

src/pixelator/pna/read_processing/pipeline.py
```diff
@@ -150,6 +179,63 @@ def _add_modifier(self, modifier: PairedEndModifier, stage: ModifierStage) -> No
...
+                    read1, read2 = out
+            else:
+                if is_r2:
+                    out = modifier(read2, info2)  # type: ignore
+                    read2 = out
+                else:
+                    out = modifier(read1, info1)
+                    read1 = out
+            if out is None:
+                break
```
johandahlberg, line 235: I'm not 100% sure I follow this. Since we are in a function that should be called for single reads I would have expected it to only deal with `SingleEndModifiers`. Perhaps we could go over this together in more detail?

tests/pna/amplicon/test_pna_cli.py
```diff
@@ -122,3 +123,31 @@ def f(**kwargs):
         ]
         cmd = runner.invoke(cli.main_cli, args)
         assert cmd.exit_code == 0
+
+
+def test_fastq_single_end(testdata_unbalanced_r12):
```
johandahlberg, line 128: Should this also have `slow` mark?

src/pixelator/pna/read_processing/pipeline.py
ptajvar: Here the name is the first out of four lines for each read in the fastq file. It looks like '@E250088191L1C001R00100000595/2'. We could stay with /2 and raise if neither /1 or /2 are present.

johandahlberg: Aha. Super. I misunderstood the context.

src/pixelator/pna/amplicon/build_amplicon.py
```diff
@@ -666,60 +708,118 @@ def _check_regions(
...
-            read1, read2, r1_regions.uei, r2_regions.uei
-        )
-        lbs1_region_qual = self._consensus_qual_lbs1(
-            read1, read2, r1_regions.lbs1, r2_regions.lbs1
-        )
-        lbs2_region_qual = self._consensus_qual_lbs2(
-            read1, read2, r1_regions.lbs2, r2_regions.lbs2
-        )
+        # Create a slice for each region in the amplicon (if it could be found)
+        r1_regions = self._scan_forward_read(read1) if read1 is not None else None
```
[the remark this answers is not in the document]
ptajvar: Sometimes a SequenceRecord evaluates to False (including when it is has an empty string sequence). So we need to explicitly check if its None.

tests/pna/amplicon/test_amplicon.py
```diff
@@ -18,3 +20,68 @@ def test_amplicon_300k(tmp_path, testdata_300k):
...
+            diff = sum([b != a for a, b in zip(p, p2)])
+            if diff < min_diff_r2:
+                min_diff_r2 = diff
+            if min_diff_r2 == 0:
+                break
+        min_diffs_r1.append(min_diff_r1)
+        min_diffs_r2.append(min_diff_r2)
+
+    assert np.mean([m > 6 for m in min_diffs_r1]) < 0.1
+    assert np.mean([m > 6 for m in min_diffs_r2]) < 0.1
```
fbdtemme, line 87: Can this be a separate utility function for the tests cases?

fbdtemme (changes requested):
I think that the current single_end / paired_end logic is too spread out. It bleeds into every part of the logic and makes it (even more) difficult to reason about. 

I think it might be better to take a step back and rethink trying to fit both cases into the single `AmpliconBuilder` class.

A `SingleEndAmpliconBuilder` and `PairedEndAmpliconBuilder` sharing a base class `AmpliconBuilderBase` and/or a bunch of utility function might work better to contain all that logic in a single place.

We can than also have a normal cutadapt SingleEndPipeline. The AmpliconPipeline was only needed to be able to model having first modifiers dealing with paired end data and the modifiers after amplicon dealing with single end data.

src/pixelator/pna/amplicon/build_amplicon.py
```diff
@@ -488,6 +489,31 @@ def _scan_reverse_read(self, read: SequenceRecord) -> AmpliconRegionSlices:
 
         return region_slices
 
+    @staticmethod
+    def _get_region_sequence(read, region_slice, is_reversed=False, as_bytearray=False):
```
fbdtemme, line 493: Might be good to add return type hints here:

```suggestion
    @staticmethod
    def _get_region_sequence(read, region_slice, is_reversed=False, as_bytearray=False) -> tuple[bytes | None, bytes | None]:
```

src/pixelator/pna/read_processing/pipeline.py
```diff
@@ -158,52 +217,65 @@ def process_reads(
...
+                )
+                total1_bp += n_bp1
+                total2_bp += n_bp2
+                final_read = self._combiner(read1, read2, info1, info2)
+
+            else:
+                read, info, n_bp = self._pre_process_single(reads)
+                total1_bp += n_bp
+                # Figure out which read to pass as None, since the combiner expects two reads.
+                if read.name.endswith("/2") or read.name.endswith("_R2"):
```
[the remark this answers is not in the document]
ptajvar: I agree about it being brittle, but this part is only reached for single-end read when there's only one fastq file provided in the amplicon command. When there are two, only the input order matters.

PR #271, opened 2025-10-06
Use pa.large_string for marker_1 / marker_2 columns

fbdtemme:
Collapsing very large PID groups where the combined marker array content exceeds 2 GiB would return an error when casting to the intermediary table format for sorting.
Explicitly use pa.large_string() instead for the table schema.

## Description

Please include a summary of the change and which issue(s) have been fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes: PNA-1486

## Type of change


ptajvar (approved):
Great! A larger sample that was originally failing now passes.

PR #285, opened 2025-10-31
Release 0.22.1

Aratz:
### Fixed

-   Panel data can now be fetched from the `.pxl` file in the denoise step
- `control` is now stored as boolean in the Panel object, instead of `yes`/`no`.
-   Fix an error in the "Linker binding site 1" region of the `pna-2` assay config file


johandahlberg (commented):
I think we should update the changelog to indicate that the `control` column will now contain boolean values.

johandahlberg (approved):

PR #289, opened 2025-11-25
Add proxiome panel v3

ptajvar:
Adding the new panel file with new AOC PIDs.


johandahlberg:
We should add a changelog entry, and I think we need to remove the default aliases, i.e. `proxiome-immuno-155` to avoid people getting the wrong panel by mistake. Before we release this we should probably coordinate with commercial and include a link to the table that they are making which shows which kit lot goes with what panel version.

johandahlberg (approved):

PR #290, opened 2025-11-28
Better adata read performance

johandahlberg:
## Description

Improved `adata` read performance when reading from many pxl files by reusing the existing connection.

## Type of change

## How Has This Been Tested?

Manually, and via the unit test suite.

### DIFF
--- a/tests/pna/pixeldataset/test_io.py
+++ b/tests/pna/pixeldataset/test_io.py
@@ -182,7 +182,8 @@ class TestPixelDataViewer:
         assert pxl_view.sample_names() == ["test_sample"]
 
     def test_read_adata_from_sample(self, pxl_view: PixelDataViewer, adata_data):
-        res = pxl_view.read_adata_from_sample("test_sample")
+        with pxl_view as connection:
+            res = pxl_view.read_adata_from_sample(connection, "test_sample")
         adata_assert_equal(res, adata_data)
 
     def test_read_adata(self, pxl_view, adata_data):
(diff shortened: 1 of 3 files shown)


tests/pna/pixeldataset/test_io.py
```diff
@@ -182,7 +182,8 @@ def test_sample_names(self, pxl_view):
         assert pxl_view.sample_names() == ["test_sample"]
 
     def test_read_adata_from_sample(self, pxl_view: PixelDataViewer, adata_data):
-        res = pxl_view.read_adata_from_sample("test_sample")
+        with pxl_view as connection:
+            res = pxl_view.read_adata_from_sample(connection, "test_sample")
```
Aratz, line 186: I'm not entirely satisfied with how the UI looks like here, i.e that this relies on the user not mixing `pxl_view` and `connection` from different files. More specifically I would prefer if `connection` was encapsulated 

Maybe it's possible to achieve something similar to how regular files work? With a context manager it would look something like:
```python
with PxlDataViewer.from_files(["/path/to/file"]) as pxl_view:
    pxl_view.read_adata_from_sample("test_sample")
```

Aratz (approved):
Great! I think it's good enough for now, but maybe we can try to improve the interface later.

Also for the record, how much faster does it get if you reuse the connection?

tests/pna/pixeldataset/test_io.py
johandahlberg: I see what you mean. I think that for all intents and purposes `PxlDataViewer` is a private class, so we might want to fix up the naming to indicate that. All actual public usage should go via the `PNAPixelDataset` class.

johandahlberg:
For 100 files and ~70k cells it goes from 120s to 16s to read the anndata.

PR #306, opened 2026-02-12
Drop umi tools dependency

johandahlberg:
## Description

We were having build issues due to UMI tools using `distutils`, see [here](https://github.com/PixelgenTechnologies/pixelator/actions/runs/21939253300/job/63362172491?pr=305). In reality we used a tiny part of the umi tools library, so I vendored one of the functions, and replaced the other with an other library.

## Type of change

Just refactoring.

## How Has This Been Tested?

Unit test suite.


ptajvar (approved):
LGTM!

PR #307, opened 2026-02-17
Ensure that additional adata columns are not removed in the denoise step

ptajvar:
Information added to adata are removed in the denoise step. This happens because marker adata is recalculated since denoising changes marker counts for each component and some of the statistics should be recalculated. The fix is to add back any columns that are missing in the new adata from the old adata.

Fixes: #PNA-2075

## Type of change

## How Has This Been Tested?

We now add a dummy column before denoising and check if it still exists after denoising.

## PR checklist:

### DIFF
--- a/src/pixelator/pna/analysis/denoise.py
+++ b/src/pixelator/pna/analysis/denoise.py
@@ -20,7 +20,7 @@ from scipy.stats import fisher_exact
 
 from pixelator.common.annotate.aggregates import call_aggregates
 from pixelator.pna.analysis_engine import PerComponentTask
-from pixelator.pna.anndata import pna_edgelist_to_anndata, add_missing_adata_info
+from pixelator.pna.anndata import add_missing_adata_info, pna_edgelist_to_anndata
 from pixelator.pna.config import pna_config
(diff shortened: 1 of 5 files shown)


src/pixelator/pna/analysis/denoise.py
```diff
@@ -30,6 +30,16 @@
...
 
 
+def _add_missing_adata_info(new_adata, old_adata):
+    missing_obs = set(old_adata.obs.columns) - set(new_adata.obs.columns)
+    missing_var = set(old_adata.var.columns) - set(new_adata.var.columns)
+
+    new_adata.obs = new_adata.obs.join(old_adata.obs[list(missing_obs)], how="left")
+    new_adata.var = new_adata.var.join(old_adata.var[list(missing_var)], how="left")
+
+    return new_adata
```
johandahlberg: I think it would make sense to put this into `anndata` module and make it public (i.e. remove the leading underscore in the name), so that we can use this in more places later.

johandahlberg (approved):
See my inline comment. Lets fix that and then merge this.

PR #310, opened 2026-02-20
Feature/pna 2090 bug filter by sample in pxl dataset

johandahlberg:
## Description

Filtering by sample was not working as expected on `PNAPixelDataset`, e.g. even when filtering you would still get data from all samples when getting the proximity dataframe. This PR addresses that.

It should be noted that this makes the `sample_names` function heavier, since it actually checks which samples are present, but if that becomes an issue we can cache it in the future. For now I prefer to keep things as simple as possible.

While working on this I found a bunch of inconsistencies in the tests, which are also addressed here, which makes the change set here quite a bigger than the main fix.

Fixes: pna-2090

## Type of change

## How Has This Been Tested?

With the test suite, and by running filtering commands manually.


ptajvar (approved):
Great fix!

PR #325, opened 2026-03-24
add celltype annotation functions

ptajvar:
Adding functions and data to enable automated annotation of PBMC cells through label transfer from a reference dataset.

Fixes: PNA-2292 and PNA-2293

## Type of change

## How Has This Been Tested?

WIP

## PR checklist:

- [X] This comment contains a description of changes (with reason).


src/pixelator/common/annotate/annotate_celltype.py
```diff
@@ -0,0 +1,283 @@
...
+from typing import List
+
+import numpy as np
+import pandas as pd
+import scanpy as sc
+from anndata import AnnData
+from sklearn.linear_model import LinearRegression
+from sklearn.neighbors import KNeighborsClassifier
+
+logging.basicConfig(level=logging.INFO)
```
johandahlberg:
```suggestion
```

No need to do any logging setup here. That should be handled higher up in the call stack.

src/pixelator/common/annotate/annotate_celltype.py
```diff
@@ -0,0 +1,283 @@
...
+        raise ValueError("min_prediction_score must be between 0 and 1.")
+
+    for ref_group in reference_groups:
+        if ref_group not in reference.obs.columns:
+            raise KeyError(f"Column '{ref_group}' not found in reference.obs")
+
+    shared_features = list(set(query.var_names).intersection(reference.var_names))
+
+    if verbose:
+        logger.info(f"Found {len(shared_features)} overlapping features.")
```
johandahlberg: Just use debug logging here instead.

src/pixelator/common/annotate/annotate_celltype.py
```diff
@@ -0,0 +1,283 @@
...
+
+    query_tmp = query[:, shared_features].copy()
+    ref_tmp = reference[:, shared_features].copy()
+
+    if not skip_normalization:
+        if verbose:
+            logger.info("Applying LogNormalize.")
```
johandahlberg: Same here.

src/pixelator/common/annotate/annotate_celltype.py
```diff
@@ -0,0 +1,283 @@
...
+        )
+    elif method == "nmf":
+        query = _seeded_nmf_annotation(
+            query=query_tmp,
+            reference=ref_tmp,
+            groups=reference_groups,
+            skip_normalization=skip_normalization,
+            min_prediction_score=min_prediction_score,
+            verbose=verbose,
+        )
```
johandahlberg:
```suggestion
        )
   else:
       raise ValueError("method must be either 'scanpy' or 'nmf'.")
   
```

johandahlberg: Or we could use a match instead here: https://docs.python.org/3/tutorial/controlflow.html#match-statements

I think that is nicer and since we only support 3.10+ it should be available on all of our supported python versions.

src/pixelator/common/annotate/annotate_celltype.py
johandahlberg: And same everywhere else this pattern is used.

src/pixelator/common/annotate/annotate_celltype.py
```diff
@@ -0,0 +1,283 @@
...
+        W[:, i] = x1 / x2
+
+    max_vals = W.max(axis=0)
+    max_vals[max_vals == 0] = 1  # Avoid division by zero
+    W = W / max_vals
+
+    return W, unique_groups
+
+
+def read_adata_from_csv(counts_path: str, meta_path: str) -> AnnData:
```
johandahlberg: Maybe this should be a private function? Or do you think that users should load data like this as well?

johandahlberg: Alternatively we could store this as a binary file anndata  (`.h5ad` file) in the resources instead and have that load with the package?

src/pixelator/common/annotate/annotate_celltype.py
```diff
@@ -0,0 +1,283 @@
...
+    """Annotate cell types transferring labels from a reference to a query.
+
+    Args:
+    query (AnnData): The unannotated target dataset.
+    reference (AnnData): The annotated reference dataset.
+    summarize_by_column (str | None): If provided, summarizes annotations to the most common label per cluster/factor in this column.
+    reference_groups (List[str]): The columns in `reference.obs` containing the ground truth labels.
+    method (str): Either 'scanpy' (PCA + KNN, mimics Seurat) or 'nmf' (Seeded Non-Negative Matrix Factorization).
+    min_prediction_score (float): Cells with prediction probabilities below this are labeled "Unknown".
+    min_cells_per_celltype (int | None): Minimum number of cells required per cell type for annotation.
```
ludvigla: It's probably best to specify that this refers to cells in the reference anndata object

ludvigla (approved):
Looks great!

src/pixelator/common/annotate/annotate_celltype.py
```diff
@@ -0,0 +1,283 @@
...
+from pixelator.common.utils import logger
+
+
+def annotate_cells(
+    query: AnnData | pd.DataFrame,
+    reference: AnnData | None = None,
+    summarize_by_column: str | None = None,
+    reference_groups: List[str] = ["celltype_l1", "celltype_l2"],
+    method: str = "nmf",
+    min_prediction_score: float = 0.2,
```
maxkarlsson: In R we use 0 as default - let's do that here as well for consistency

src/pixelator/common/annotate/annotate_celltype.py
```diff
@@ -0,0 +1,283 @@
...
+            query = _scanpy_annotation(
+                query=query_tmp,
+                reference=ref_tmp,
+                groups=reference_groups,
+                min_prediction_score=min_prediction_score,
+            )
+        case "nmf":
+            if n_cells_per_group is None:
+                n_cells_per_group = 50
+            if min_cells_per_celltype is None:
```
maxkarlsson, line 130: This is a difference from the R implementation, which doesn't have a minimum 😮

maxkarlsson: Aa sorry, no R does too - it just isn't propagated to the main function

src/pixelator/common/annotate/annotate_celltype.py
```diff
@@ -0,0 +1,283 @@
...
+                )
+            query = _scanpy_annotation(
+                query=query_tmp,
+                reference=ref_tmp,
+                groups=reference_groups,
+                min_prediction_score=min_prediction_score,
+            )
+        case "nmf":
+            if n_cells_per_group is None:
+                n_cells_per_group = 50
```
maxkarlsson: This parameter is default 100 in the R implementation

src/pixelator/common/annotate/annotate_celltype.py
```diff
@@ -0,0 +1,283 @@
...
+
+        W, unique_groups = _get_w_matrix(
+            ref_data=ref_data.T,
+            groups=labels,
+            n_cells_per_group=n_cells_per_group,
+            min_cells_per_celltype=min_cells_per_celltype,
+            seed=seed,
+        )
+
+        nnls_solver = LinearRegression(fit_intercept=False, positive=True)
```
maxkarlsson, line 226: A difference from the R implementation is that we are not doing any L1 regularization. I did some searching and this _should_ be mathematically equivalent to the R implementation:
```
# Inside _seeded_nmf_annotation, in the per-group loop, after W is built:
RCPPML_L1 = 0.01
n_features = W.shape[0]
alpha_equiv = RCPPML_L1 / n_features
nnls_solver = Lasso(
    alpha=alpha_equiv,
    positive=True,
    fit_intercept=False,
    max_iter=100_000,   # RcppML iterates to tight convergence; give sklearn room
    tol=1e-8,           # match RcppML's default convergence tolerance more closely
    selection="cyclic", # RcppML uses cyclic coordinate descent
)
nnls_solver.fit(W, target_data.T)
H = nnls_solver.coef_  # Shape: (Cells x Groups)
```

maxkarlsson (commented):
I think it looks very good! I found one difference between the implementation in the L1 regularization in the R implementation and added a suggestion for something we could try.

johandahlberg:
@ptajvar This is missing a changelog entry. Could you add that in a separate RP?

PR #329, opened 2026-04-01
Better samplesheet validation for hash indexes

johandahlberg:
## Description

Add better samplesheet validation for the samplesheet step.

Fixes: pna-2487

## Type of change

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce it when relevant.


Aratz (approved):
Looks good!

However, maybe we should consider doing the validation upfront, at the beginning of the pipeline?

If this validation is only done at the sample calling step and fails, I'm afraid updating the samplesheet would trigger a full rerun (depending how nextflow detects changes in the samplesheet, might be worth to try it out first).

johandahlberg:
Yes. I agree. That would be preferable. I think maybe we should add a pre-flight step - but I think that is out of scope here.

PR #340, opened 2026-04-15
Release 0.25.0

Aratz:
### Added

- Add sample calling functionality
- Add FLAG and FMC63 v3 panels

### Changed
- Change default value for parameter axis of function clr_transformation to be 1.

### Fixed

- Fix logging for log1p_transformation and clr_transformation to state that the stats correspond to components and not nodes.
- Fix bug in unpacking 4bit packed sequences.
- Fix bug in sample calling that removed the ending of antibodies that ended with `-<digit>`.


johandahlberg (approved):
I realized we are missing an entry in the changelog about the refactor of the pixeldataset - should we maybe add that? You dediced.

Otherwise this looks good to go.

Aratz:
Sounds like a good idea, I'll add it :+1:

PR #352, opened 2026-04-23
Pna 2591 - Add adaptive_core_expansion and node_pls algorithms for cell denoising

ludvigla:
## Description

This PR adds two new algorithms (described in detail in pixelatorR-internal) for graph denoising. Both tools take a `Graph` object as input, including a networkx graph and a count matrix, and outputs the same object with additional node metrics which can be used for downstream filtering.

- adaptive_core_expansion: ACE utilizes topology-aware graph partitioning. It identifies a high-density k-core "seed" and iteratively expands it using transition probabilities to recruit peripheral nodes. The final partition is determined by maximizing the Bray-Curtis dissimilarity between the "core" and "periphery" layers.
- node_pls: This method models the statistical relationship between node neighborhood abundance and graph coreness. By assigning each node calculated "denoising scores" we can apply targeted filtering.

This PR also adds additional options to the `denoise` cli, enabling ACE and PLS denoising.

Fixes: PNA-2591, and PNA-2671

## Type of change

## How Has This Been Tested?

Added tests in `tests/common/test_adaptive_core_expansion.py` and `tests/common/test_node_pls.py`.

## PR checklist:


src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,218 @@
...
+
+    k_cores[k_cores > max_k] = max_k
+
+    # Compute transition probability matrix P from the adjacency matrix
+    A = cg.get_adjacency_sparse(node_ordering=node_list)
+    A = A + sp.diags_array([1] * A.shape[0], format="csr", dtype=None)
+
+    row_sums = np.ravel(A.sum(axis=1))
+    D_inv = sp.diags_array(1 / row_sums, format="csr")
+    P = D_inv @ A
```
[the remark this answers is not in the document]
ptajvar: We can bump it to 1.11 when this was added.

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,218 @@
...
+    :param select_LCC: Restricts the initial seed to the Largest Connected Component.
+    :param verbose: Whether to print progress alerts.
+    :return: The original graph object with an additional `partition` node attribute ("high" or "low").
+    """
+    assert 1 <= k <= 6
+    assert 2 <= max_k_core <= 10
+    assert all(0 <= t <= 1 for t in binding_thresholds)
+    assert 1 <= max_iter <= 1000
+    assert 0 <= min_seed_pct <= 1
+    assert 0 <= nodes_to_move_threshold <= 1000
```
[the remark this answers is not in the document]
ptajvar: We're using asserts in almost all of the tests right now so let's keep it as it is for now.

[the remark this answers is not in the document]
johandahlberg: It's actually a big difference between the tests and the production code here. The copilot suggestion above is actually correct.

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,218 @@
...
+    k: int = 3,
+    max_k_core: int = 4,
+    binding_thresholds: Sequence[float] = (0.5, 0.475, 0.45, 0.425, 0.4, 0.375, 0.35, 0.325, 0.3),
+    max_iter: int = 200,
+    min_seed_pct: float = 0.1,
+    nodes_to_move_threshold: int = 10,
+    select_LCC: bool = True,
+    verbose: bool = True,
+) -> Graph:
+    """Performs a topology-aware graph partitioning by identifying a high-density
```
johandahlberg: I realize that we are not 100% consistent on this, but we are slowly moving all the doc strings over to google style doc strings (see an example here: https://github.com/PixelgenTechnologies/pixelator/blob/b52d35fea500539e578146b1465c86f12af0a983/src/pixelator/pna/pixeldataset/download.py#L84)

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,218 @@
...
+    P_step.setdiag(0)
+
+    row_sums = np.ravel(P_step.sum(axis=1))
+    D_inv = sp.diags_array(1 / row_sums, format="csr")
+    P_step = D_inv @ P_step
+
+    if verbose:
+        logger.info(
+            f"Seed 'high' core layer contains {np.mean(k_cores == max_k)*100:.2f}% of all nodes."
+        )
```
johandahlberg: No need to use the `if verbose: ...` pattern here. Put them as `logger.debug(...)` messages and then the logging will properly handle when to output what.

src/pixelator/common/graph/node_pls.py
```diff
@@ -0,0 +1,214 @@
...
+def _residualize_matrix(X: np.ndarray, model_mat: np.ndarray) -> np.ndarray:
+    """Residualize matrix X against model_mat using OLS.
+
+    :param X: Matrix to residualize (n_samples, n_features)
+    :param model_mat: Covariates matrix (n_samples, n_covariates)
+    :return: Residualized matrix
+    """
+    # Solve M @ B = X for B
+    B, _, _, _ = np.linalg.lstsq(model_mat, X, rcond=None)
+    return X - model_mat @ B
```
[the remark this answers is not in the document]
maxkarlsson: My opinion is tahat as long as it is possible to use any model matrix that the user wants I don't see any reason why we should force an intercept here. I think residualization is something that you should do with explicit intent, so enforcing an intercept seems to just make it less flexible and possibly confusing. If the user don't want an intercept they should be allowed to do so at their own peril.

src/pixelator/common/graph/node_pls.py
```diff
@@ -0,0 +1,214 @@
...
+    use_weights: bool = False,
+    min_weight: float = 0.0,
+    normalization: Literal["L1", "CLR", "LogNormalize", "none"] = "L1",
+    scale: bool = True,
+    model_mat: Optional[np.ndarray] = None,
+) -> pd.DataFrame:
+    """Create a node abundance matrix for a Graph object.
+
+    The node abundance matrix is generated by expanding the counts matrix of the Graph
+    to include neighborhood counts up to `k` steps away.
```
maxkarlsson: The docstrings are a lot less verbose than in `pixelatorR` - is that intentional? What do we prefer?

src/pixelator/common/graph/node_pls.py
```diff
@@ -0,0 +1,214 @@
...
+    y_not_in_counts = [y for y in y_vars if y not in counts.columns]
+
+    # Extract Y from expanded matrix if present in counts
+    Y_data = {}
+    if y_in_counts:
+        for y in y_in_counts:
+            Y_data[y] = X_expanded[y].values
+        # Drop response variables from X to avoid data leakage
+        X_expanded = X_expanded.drop(columns=y_in_counts)
+
```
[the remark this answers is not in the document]
maxkarlsson: This seems like a wild edge case. I don't think we need to consider that (?)

maxkarlsson (commented):
As far as I can see it looks good, but perhaps it would be best if we try to calculate som matrices and PLS models here and compare them with R?

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,218 @@
...
+            bc_score = 0.0
+        else:
+            high_mask = current_partition == 1
+            low_mask = ~high_mask
+
+            # Use pandas direct indexing since node_marker_counts indices ordered same as node_list
+            x = normalize_counts(counts[high_mask].sum(axis=0).values)
+            y = normalize_counts(counts[low_mask].sum(axis=0).values)
+
+            bc_score = float(np.sum(np.abs(x - y)) / np.sum(x + y))
```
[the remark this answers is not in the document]
ludvigla: This should never happen unless node protein counts have been set to 0 for some reason.

src/pixelator/common/graph/node_pls.py
ludvigla: I have added more content from the R docs now!

ludvigla:
I have now added an additional parameter `min_allowed_nodes_pct` to `adaptive_core_expansion` (default 0.8). If the best partition (with highest Bray-Curtis score) includes a "high" core with  less then `min_allowed_nodes_pct`, the partition with the highest Bray-Curtis score that falls above the `min_allowed_nodes_pct` threshold is used instead.

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,342 @@
...
+        else:
+            logger.warning(
+                f"Found no partition with at least {min_allowed_nodes_pct * 100:.2f}% of "
+                f"nodes in the 'high' core layer. Setting all nodes as 'high'."
+            )
+            best_candidate = None
+
+    if best_candidate is not None:
+        best_partition_vec = best_candidate["partition"]
+        logger.info(
```
ptajvar: We should probably change this to logger.debug, otherwise it is always printed.

src/pixelator/pna/analysis/denoise.py
```diff
@@ -183,19 +187,200 @@ def get_stranded_nodes(component: PNAGraph, nodes_to_remove: list = []) -> list:
...
+            model_mat=model_mat,
+        )
+        X_pred = _create_node_neighborhood_abundance_matrix(
+            component,
+            k=pred_k,
+            use_weights=use_weights,
+            normalization=normalization,
+            model_mat=model_mat,
+        )
+        scores = pls_model.transform(X_pred.values)
```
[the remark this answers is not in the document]
ludvigla: What do you think about this comment @maxkarlsson ?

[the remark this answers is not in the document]
maxkarlsson: Yeah, we do see that the k2 model scores and k1 predicted scores differ somewhat and they don't have the same correlation to the Y. However, since the performance seems better with k1 prediction I would still opt for it :)

src/pixelator/pna/cli/denoise.py
```diff
@@ -79,16 +112,103 @@
...
+    show_default=True,
+    help="Keep PLS components whose Pearson correlation with coreness exceeds this.",
+)
+@click.option(
+    "--pls-score-threshold",
+    default=-3.0,
+    type=float,
+    show_default=True,
+    help="All retained PLS score columns must exceed this value when --run-pls-denoising is set.",
+)
```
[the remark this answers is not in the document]
ptajvar: Now added a CLI test for this.

src/pixelator/pna/analysis/denoise.py
```diff
@@ -183,19 +187,200 @@ def get_stranded_nodes(component: PNAGraph, nodes_to_remove: list = []) -> list:
...
+
+def _nodes_outside_largest_cc_after_pls_scores(
+    raw: nx.Graph,
+    node_index: pd.Index,
+    passing_mask: np.ndarray,
+) -> list:
+    """Return nodes to drop: fail score filter or lie outside the largest CC among passers."""
+    keep_candidates = [node_index[i] for i in range(len(node_index)) if passing_mask[i]]
+    if not keep_candidates:
+        return list(raw.nodes)
```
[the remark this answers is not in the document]
ptajvar: This is the intended outcome. We should not have component information about a completely removed component in adata.

maxkarlsson (approved):
I had a look at the PLS stuff and I think it looks great! I tried to see if I can find any differences to the R implementation but I cannot 👍

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,351 @@
...
+class _PartitionCandidate:
+    partition: np.ndarray
+    bc_score: float
+    nodes_pct: float
+    binding_threshold: float
+
+
+def normalize_counts(x):
+    """Normalize counts using log-mean-exp normalization."""
+    # counts_mat: shape (n_nodes, n_markers)
```
johandahlberg:
```suggestion
```

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,351 @@
...
+
+from dataclasses import dataclass
+from typing import Sequence
+
+import networkx as nx
+import numpy as np
+import scipy.sparse as sp
+
+from pixelator.common.graph import Graph
+from pixelator.common.graph.backends.implementations._networkx import _mat_pow
```
johandahlberg: Not super critical, but since this method leaks hear, and it does not seem to be directly tied to the networkx graph implementation I think that perhaps we should create a new module called something like: `pixelator.common.graph.math` and put it there.

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,351 @@
...
+        0.375,
+        0.35,
+        0.325,
+        0.3,
+    ),
+    max_iter: int = 200,
+    min_seed_pct: float = 0.1,
+    nodes_to_move_threshold: int = 10,
+    min_allowed_nodes_pct: float = 0.8,
+    select_LCC: bool = True,
```
johandahlberg: Nitpicking, but I think we should use lower case here instead of mixing, i.e. `select_lcc` for consistency.

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,351 @@
...
+    """Normalize counts using log-mean-exp normalization."""
+    # counts_mat: shape (n_nodes, n_markers)
+    x_log1p = np.log1p(x)
+    x_mean = np.mean(x_log1p)
+    norm = np.exp(x_mean)
+    return np.log1p(x / norm)
+
+
+def adaptive_core_expansion(
+    cg: Graph,
```
johandahlberg: For consistency with other methods I'd suggest calling this either `graph` or `g`.

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,351 @@
...
+def _validate_ace_parameters(
+    k: int,
+    max_k_core: int,
+    binding_thresholds: Sequence[float],
+    max_iter: int,
+    min_seed_pct: float,
+    nodes_to_move_threshold: int,
+    min_allowed_nodes_pct: float,
+) -> bool:
+    if not 1 <= k <= 6:
```
johandahlberg, line 322: Since this just raises, and the value is not used, I would probably make the return type here `-> None` instead and just exit the function without returning at the end if no exceptions were triggered. This feels slightly more pythonic to me.

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,351 @@
...
+    k_cores_dict = nx.core_number(raw_graph)
+    k_cores = np.array([k_cores_dict[n] for n in node_list])
+    max_k = k_cores.max()
+
+    if max_k == 1:
+        raise ValueError(
+            "The graph does not contain any k-core layers above 1. The graph has too low connectivity."
+        )
+
+    if max_k > max_k_core:
```
johandahlberg, line 143:
```suggestion
    # Cap max k-cores to max_k_core if higher k-core layers are identified.
    if max_k > max_k_core:
```
Perhaps some comment on what is going on here to make the code read a bit easier.

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,351 @@
...
+        k_cores[k_cores > max_k_core] = max_k_core
+        max_k = max_k_core
+
+    pct_k_max = np.mean(k_cores == max_k)
+    while pct_k_max < min_seed_pct and max_k > 1:
+        max_k -= 1
+        pct_k_max = np.mean(k_cores == max_k)
```
johandahlberg, line 151:
```suggestion
    # Expand the k-core layer if it is smaller than the minimum seed percentage.
    pct_k_max = np.mean(k_cores == max_k)
    while pct_k_max < min_seed_pct and max_k > 1:
        max_k -= 1
        pct_k_max = np.mean(k_cores == max_k)
```

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,351 @@
...
+    # Compute transition probability matrix P from the adjacency matrix
+    A = cg.get_adjacency_sparse(node_ordering=node_list)
+    A = A + sp.diags_array([1] * A.shape[0], format="csr", dtype=None)
+
+    row_sums = np.ravel(A.sum(axis=1))
+    D_inv = sp.diags_array(1 / row_sums, format="csr")
+    P = D_inv @ A
+
+    min_weight = 1e-5  # To avoid having the sparse matrix grow too dense
+    P_step = _mat_pow(P, k, prune_threshold=min_weight).T
```
johandahlberg, line 173: We have almost identical code for this in the `_prob_edge_weights` function in `src/pixelator/common/graph/backends/implementations/_networkx.py` file. Perhaps this is something we should also break out into a reusable matrix-math-on-graph module?

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,351 @@
...
+    P_step.setdiag(0)
+
+    row_sums = np.ravel(P_step.sum(axis=1))
+    D_inv = sp.diags_array(1 / row_sums, format="csr")
+    P_step = D_inv @ P_step
+
+    if select_LCC:
+        seed_nodes = [node_list[i] for i, k_val in enumerate(k_cores) if k_val == max_k]
+        subgraph = raw_graph.subgraph(seed_nodes)
+        components = list(nx.connected_components(subgraph))
```
johandahlberg, line 184: Since we have the adjacency matrix, perhaps we could use the scipy sparse matrix connected components here instead, as it might be faster: https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csgraph.connected_components.html

Not sure about this though, so do what you will with it. This should also work fine.

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,351 @@
...
+            _PartitionCandidate(
+                partition=current_partition.copy(),
+                bc_score=bc_score,
+                nodes_pct=float(num_high / len(current_partition)),
+                binding_threshold=binding_threshold,
+            )
+        )
+
+    best_idx = int(np.argmax([p.bc_score for p in partitions]))
+    best_candidate: _PartitionCandidate | None = partitions[best_idx]
```
johandahlberg, line 241: I think this can be simplified into something like:
```
best_calidatate = max(partitions, key=lambda p: p.bc_score)
```

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,351 @@
...
+        best_partition_vec = np.ones(len(node_list), dtype=int)
+        logger.debug("No suitable partition found. All nodes set to 'high'.")
+
+    pixel_type_dict = {
+        node_list[i]: ("high" if best_partition_vec[i] == 1 else "low")
+        for i in range(len(node_list))
+    }
+    nx.set_node_attributes(raw_graph, pixel_type_dict, "partition")
+
+    return cg
```
johandahlberg: Would it perhaps make more sense to return the partitions directly here instead of setting them on the graph? In general I think that is how the partitioning functions in networkx works. It also does not mutate the input data with out the knowledge of the caller, which I think it good as a general principle.

src/pixelator/common/graph/adaptive_core_expansion.py
```diff
@@ -0,0 +1,351 @@
...
+    nodes_to_move_threshold: int,
+    binding_threshold: float,
+) -> np.ndarray:
+    for i in range(max_iter):
+        tp_total = P @ partition
+
+        to_move_mask = (partition == 0) & (tp_total > binding_threshold)
+        num_to_move = to_move_mask.sum()
+
+        if num_to_move < nodes_to_move_threshold:
```
johandahlberg, line 300: Do  we also want to log if we reach the `max_iter`? I don't know if this is useful here or not.

src/pixelator/common/graph/node_pls.py
```diff
@@ -0,0 +1,288 @@
...
+    if scale:
+        std = np.std(X, axis=0)
+        std[std == 0] = 1
+        X = (X - np.mean(X, axis=0)) / std
+
+    return pd.DataFrame(X, index=counts.index, columns=counts.columns)
+
+
+def node_pls(
+    cg: Graph,
```
johandahlberg: Same comment on the name as on the first other main function, i.e. perhaps we want to use `g` or `graph` here for consistency.

src/pixelator/common/graph/node_pls.py
```diff
@@ -0,0 +1,288 @@
...
+
+    Raises:
+        ValueError: If the number of rows in model_mat does not match the
+            number of nodes in the graph.
+
+    """
+    counts = cg.node_marker_counts
+    X = counts.values.astype(np.float64)
+
+    if k > 0:
```
johandahlberg, line 95: Not super important, but perhaps we want to first raise a `ValueError` if k < 0?

src/pixelator/common/graph/node_pls.py
```diff
@@ -0,0 +1,288 @@
...
+
+            if not use_weights:
+                # Set all non-zero entries to 1 to get unweighted neighborhood counts
+                W.data = np.ones_like(W.data)
+
+        else:
+            W = A
+
+        # Neighborhood abundance is W @ X
+        X = (W @ X).astype(np.float64)
```
johandahlberg, line 118: I think that for readability/testabilty this can probably be broken out into a sub-function like `compute_neighborhood_abundance`.

src/pixelator/common/graph/node_pls.py
```diff
@@ -0,0 +1,288 @@
...
+
+def node_pls(
+    cg: Graph,
+    y_vars: Union[str, List[str]],
+    x_vars: Optional[List[str]] = None,
+    k: int = 2,
+    use_weights: bool = False,
+    min_weight: float = 0.0,
+    ncomp: int = 1,
+    normalization: Literal["L1", "CLR", "LogNormalize", "none"] = "L1",
```
johandahlberg, line 160: I think this should be `Literal["L1", "CLR", "LogNormalize", "none"] | None` and then we should handle the `None` case specifically. That feels more pythonic to me.

src/pixelator/common/graph/node_pls.py
```diff
@@ -0,0 +1,288 @@
...
+        std = np.std(X, axis=0)
+        std[std == 0] = 1
+        X = (X - np.mean(X, axis=0)) / std
+
+    return pd.DataFrame(X, index=counts.index, columns=counts.columns)
+
+
+def node_pls(
+    cg: Graph,
+    y_vars: Union[str, List[str]],
```
johandahlberg, line 154: Since python 3.10 the standard for type-hinting in this scenario is: `str | List[str]`. https://peps.python.org/pep-0604/

src/pixelator/common/graph/node_pls.py
```diff
@@ -0,0 +1,288 @@
...
+    x_vars: Optional[List[str]] = None,
```
johandahlberg, line 155: And instead of optional we can use `List[str] | None`.

src/pixelator/common/graph/node_pls.py
```diff
@@ -0,0 +1,288 @@
...
+    k: int = 2,
+    use_weights: bool = False,
+    min_weight: float = 0.0,
+    ncomp: int = 1,
+    normalization: Literal["L1", "CLR", "LogNormalize", "none"] = "L1",
+    scale: bool = True,
+    model_mat: Optional[np.ndarray] = None,
```
johandahlberg, line 162: Same here, it should use `np.ndarray | None`

src/pixelator/common/graph/node_pls.py
```diff
@@ -0,0 +1,288 @@
...
+            before being subset out for Y.
+        x_vars: A list of variable names to use as predictors (X). These
+            should be column names from the counts matrix of the Graph. If None,
+            all columns from the counts matrix will be used as predictors.
+            Defaults to None.
+        k: An integer value specifying the maximum steps from each node to
+            expand the neighborhood for creating the predictor matrix X. If `k > 0`,
+            the counts will be expanded to include neighborhood counts up to `k`
+            steps away using the adjacency matrix of the Graph. Defaults to 2.
+        use_weights: Logical indicating whether to use edge weights when
```
johandahlberg:
```suggestion
        use_weights: Boolean indicating whether to use edge weights when
```

src/pixelator/pna/analysis/denoise.py
```diff
@@ -183,19 +187,210 @@ def get_stranded_nodes(component: PNAGraph, nodes_to_remove: list = []) -> list:
...
+
+
+def denoise_pls(
+    component: PNAGraph,
+    *,
+    ncomp: int = 5,
+    model_k: int = 2,
+    pred_k: int = 1,
+    use_weights: bool = True,
+    normalization: Literal["L1", "CLR", "LogNormalize", "none"] = "L1",
```
johandahlberg, line 272: Same comment as on other function. Should use `Literal["L1", ...] | None`

src/pixelator/pna/analysis/denoise.py
```diff
@@ -183,19 +187,210 @@ def get_stranded_nodes(component: PNAGraph, nodes_to_remove: list = []) -> list:
...
+        component: Component graph (mutates ``coreness`` on nodes temporarily).
+        ncomp: Requested PLS components (capped by sample size and feature count).
+        model_k: Neighborhood steps for fitting X.
+        pred_k: Neighborhood steps for prediction / scores.
+        use_weights: Use edge weights in neighborhood expansion.
+        normalization: Neighborhood matrix normalization.
+        residualize: If True, residualize X against a ``pixel_type`` design matrix.
+        pls_component_p_threshold: Per-component Pearson test vs coreness.
+        min_pls_coreness_correlation: Minimum positive correlation.
+        pls_score_threshold: All selected score columns must exceed this.
```
johandahlberg, line 296: In the other docs we document which value is the default. Perhaps we should do the same here?

src/pixelator/pna/analysis/denoise.py
```diff
@@ -183,19 +187,210 @@ def get_stranded_nodes(component: PNAGraph, nodes_to_remove: list = []) -> list:
...
+    nx.set_node_attributes(component.raw, coreness_series.to_dict(), "coreness")
+    model_mat: Optional[np.ndarray] = None
+    if residualize:
+        model_mat = _pixel_type_design_matrix(component, idx)
+
+    def _cleanup_coreness() -> None:
+        for n in list(component.raw.nodes):
+            data = component.raw.nodes[n]
+            if isinstance(data, dict) and "coreness" in data:
+                del data["coreness"]
```
johandahlberg, line 324: Since we need to do the clean-up I get a gut feeling that perhaps it would better to keep this data out of the graph and/or only apply it when it needed. Instead of setting and removing it.

src/pixelator/common/graph/node_pls.py
```diff
@@ -0,0 +1,288 @@
...
+    :param X: Matrix to residualize (n_samples, n_features)
+    :param model_mat: Covariates matrix (n_samples, n_covariates)
+    :return: Residualized matrix
+    """
+    # Solve M @ B = X for B
+    B, _, _, _ = np.linalg.lstsq(model_mat, X, rcond=None)
+    return X - model_mat @ B
+
+
+def _create_node_neighborhood_abundance_matrix(
```
johandahlberg, line 34: Since this is imported in other modules it should not be prefixed by `_`, marking it as private.

src/pixelator/pna/analysis/denoise.py
```diff
@@ -183,19 +187,210 @@ def get_stranded_nodes(component: PNAGraph, nodes_to_remove: list = []) -> list:
...
+        # (higher score = more core-like). This is required to match the R implementation
+        # (rpls) where scores are oriented by positive correlation with coreness.
+        if r_model < 0:
+            r = -r
+            scores[:, j] = -scores[:, j]
+        if r > min_pls_coreness_correlation and p < pls_component_p_threshold:
+            comps_to_use.append(j)
+
+    if not comps_to_use:
+        return []
```
johandahlberg, line 374: This feel like a candidate for a subfunction to me.

src/pixelator/pna/analysis/denoise.py
```diff
@@ -262,78 +457,145 @@ def write_denoised_edgelist(
...
+        max_k_core: int = 4,
+        max_iter: int = 200,
+        min_seed_pct: float = 0.1,
+        nodes_to_move_threshold: int = 10,
+        ace_select_lcc: bool = True,
+        pls_ncomp: int = 5,
+        pls_model_k: int = 2,
+        pls_pred_k: int = 1,
+        pls_use_weights: bool = True,
+        pls_normalization: Literal["L1", "CLR", "LogNormalize", "none"] = "L1",
```
johandahlberg, line 484: Since this comes up in so many places, perhaps it would be good to break it out into a type alias?

src/pixelator/pna/analysis/denoise.py
```diff
@@ -262,78 +457,145 @@ def write_denoised_edgelist(
...
+        pls_residualize: bool = False,
+        pls_component_p_threshold: float = 0.01,
+        min_pls_coreness_correlation: float = 0.0,
+        pls_score_threshold: float = -3.0,
```
johandahlberg, line 488: Since this explodes in the number of arguments, perhaps it would be nice to break them into a number of data classes, and then have a type signature here like `pls_conf: PLSConfig | None = None` and then check when we are deciding what to run like so:

```
if pls_conf:
    ...
```

I feel like it could make it a bit clearer what parameters belong to what method, and could clean-up the code a bit.

tests/pna/data/PNA055_Sample07_S7.layout.pxl
```diff
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:f97f8fb13790711f0df6ab1f654ece94c994377b740488b34f9ba929039fb1e3
+oid sha256:9ea12d7216921248164d412b406aff86e5be0744bd87612b3ef19163e21b3e04
```
ptajvar, line 2: What has changed here is the metadata pxl file metadata that switches panel name from proxiome-immuno-155 to proxiome-v1-immuno-155-v1.0

src/pixelator/pna/analysis/denoise.py
```diff
@@ -356,20 +618,82 @@ def add_to_pixel_file(self, data: pd.DataFrame, pxl_file_target: PxlFile):
...
+
+                for col in denoise_info.columns:
+                    if col in adata.obs.columns:
+                        adata.obs = adata.obs.drop(columns=[col])
                 adata.obs = adata.obs.join(denoise_info, how="left")
 
                 writer.write_adata(adata)
+
+
+def _collect_denoise_summary_info(
```
johandahlberg, line 639: This is probably good enough for now. If this becomes more complicated that this we should perhaps consider developing something a bit more custom, since this feels like it could be easy to forget about if one updates any of the methods/metrics.

src/pixelator/pna/analysis/report.py
```diff
@@ -44,5 +44,3 @@ class DenoiseSampleReport(SampleReport):
     report_type: str = "denoise"
     number_of_umis_removed: int | None
     ratio_of_umis_removed: float | None
-    number_of_disqualified_components: int | None
-    ratio_of_disqualified_components: float | None
```
johandahlberg, line 48: How come this is removed?

src/pixelator/pna/cli/denoise.py
```diff
@@ -43,6 +43,39 @@
     is_flag=True,
     help="Run the denoise step to remove markers that are over-expressed in the one-core layer of a component.",
 )
+@click.option(
+    "--run-ace-denoising",
+    required=False,
+    is_flag=True,
+    help=(
+        "Remove nodes in the peripheral-like ('low') partition from adaptive core expansion (ACE). "
```
johandahlberg:
```suggestion
        "Remove nodes in the peripheral-like partition from adaptive core expansion (ACE). "
```

It think the low part here felt like an unnecessary level of detail.

tests/common/test_adaptive_core_expansion.py
```diff
@@ -0,0 +1,150 @@
...
+        "low": partitions.count("low"),
+    }
+    # If no partition meets 99.9%, it should set all to "high"
+    assert partition_counts["low"] == 0
+    assert partition_counts["high"] == len(res.raw.nodes())
+    assert "Found no partition with at least 99.90%" in caplog.text
+
+
+@pytest.mark.slow
+def test_adaptive_core_expansion_isotype_reduction(graph_from_pxl):
```
johandahlberg:
```suggestion
def test_adaptive_core_expansion_isotype_reduction():
```

tests/pna/denoise/test_denoise.py
```diff
@@ -254,27 +431,32 @@ def test_get_stranded_nodes():
...
-    components = pna_pxl_dataset.adata().obs.index
+    components = denoise_pxl_dataset.adata().obs.index
 
     for comp in components:
         comp_graph = PNAGraph.from_edgelist(
-            pna_pxl_dataset.filter(components=[comp]).edgelist().to_polars().lazy()
+            denoise_pxl_dataset.filter(components=[comp]).edgelist().to_polars().lazy()
         )
         nodes_to_be_removed = denoise_one_core_layer(comp_graph)
+        if nodes_to_be_removed == [None]:
```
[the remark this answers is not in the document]
johandahlberg: @ptajvar perhaps check if this is correct or not?

tests/pna/denoise/test_denoise_cli.py
```diff
@@ -2,16 +2,23 @@
...
 import pytest
 from click.testing import CliRunner
 
 from pixelator import cli
+from pixelator.pna.config import pna_config
+from pixelator.pna.config.panel import load_antibody_panel
 from pixelator.pna.pixeldataset import read
 
+REFERENCE_COMPONENT = "57129a8b0fff38c6"
+ACE_LOW_NODE_COUNT = 5436
```
johandahlberg: Since these are only used once I think it makes more sense to just hard code them in the test where they are used.

johandahlberg (commented):
This looks great. I have focused the review on readability and pythonisity (if that is a word). I have made inline comments where I think updates might be useful, but there is nothing here that blocks merging.

A high level comment is that I think the `adaptive_core_expansion` function could benefit from being split into multiple smaller subfunctions to improve the testability, e.g. one for each conceptual stage:
 - `find_k_core_seed`
 - `create_transition_probablity_matrix`
 - `select_largest_connected_component`
 - `find_partition` (called once per `binding_threhold`)
 - `select_best_parition_candidate`

Great work the whole team of @ptajvar, @maxkarlsson and @ludvigla in getting this over the finishing line so quickly!

src/pixelator/pna/analysis/report.py
ptajvar: A component would have been disqualified for having a k=1 layer that was larger than a given threshold. That only made sense for core-one denoising. Now that we multiple methods, a component may skip one denoising method but be denoised by another method. So instead of this, we store how many nodes are marked by each method to be removed from each component in the anndata.

PR #364, opened 2026-05-19
Fix up denoise normalization and fix changelog

johandahlberg:
## Description

Fix the denoise normalization option to  actually use `None` and update the changelog with entries about missing changes.

## Type of change

Small fix-ups

## How Has This Been Tested?

With the unit tests, and by manually testing the `None` normalization option to denoise.


ptajvar (approved):
LGTM!

PR #390, opened 2026-06-10
Manage memory usage when building the anndata instance

johandahlberg:


ptajvar:
Marking as ready to get a container for trying in the pipeline. Still WIP.

ptajvar (approved):
LGTM. The memory improvement in denoise is also confirmed for a few 8K-cell samples.

ludvigla (approved):
Look great Pouria. As I understand it, the new solution is quite a bit slower but uses significantly less memory?

ptajvar:
> Look great Pouria. As I understand it, the new solution is quite a bit slower but uses significantly less memory?

Surprisingly, it was even faster for the 8K-cell samples, at least on my laptop (70s vs 200s). But it is a bit slower on smaller samples.

PR #408, opened 2026-07-02
Docs deploy (new)

karlmoresco:
## Description

Implemented deployment workflow for API docs with manual and on-release triggering, and support for versioning with older versions kept. Did minor changes to the Sphinx configurations, fixed som formatting issues and added a logo to the page (light). These additions were necessary in order to publish API docs for Pixelator.

Fixes: #PNA-2145 (https://linear.app/pixelgen-technologies/issue/PNA-2145/setup-api-docs-for-pixelator)

## Type of change

## How Has This Been Tested?

No tests were run manually.

## PR checklist:


.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,116 @@
...
+          fetch-depth: 0
+          fetch-tags: true
+      - uses: astral-sh/setup-uv@v6
+        with:
+          python-version: "3.13"
+      - run: uv sync --group docs
+      - name: Build docs
+        env:
+          DOCS_VERSION: ${{ steps.ver.outputs.version }}
+        run: uv run sphinx-build -b html docs docs/_build/html --keep-going -n
```
Aratz, line 49:
```suggestion
        run: uv run sphinx-build -b html docs docs/_build/html --fail-on-warning -n
```
Here I would actually fail this on warning, so that we don't miss smaller bugs that would otherwise be hard to pick up (e.g. broken links).

Right now, there are quite many such warnings, so you'll need to address these too 😉 

All this can be in a second PR though, if you want to keep this one focused.

.github/workflows/deploy-docs.yml
Aratz, line 1: One issue with building the docs only on release is that if a PR breaks the docs by mistake, it will only be picked up at release time.

It would be nice to split this workflow into two jobs instead: a build job and a deploy job.

The build job should run on PRs and releases, while the deploy job runs only on releases. This way, if the docs break, we will know immediately and avoid merging the buggy code into the dev branch.

docs/conf.py
```diff
@@ -86,10 +90,18 @@
...
+        "version_match": _docs_version,
+    },
+    "logo": {
+        "image_light": "_static/pixelator.svg",
+    },
 }
 
 html_static_path = ["_static"]
 html_css_files = ["custom.css"]
+html_extra_path = ["switcher.json"]
```
Aratz: Sphinx complains this file doesn't exist at build time, even in the github action, is this line really necessary?

https://github.com/karlmoresco/pixelator/actions/runs/28588771292/job/84766822612#step:6:12

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,116 @@
...
+          
+      - name: Checkout gh-pages (for existing switcher.json)
+        uses: actions/checkout@v4
+        continue-on-error: true
+        with:
+          ref: gh-pages
+          path: gh-pages
+          fetch-depth: 1
+
+      - name: Update switcher.json (append if missing)
```
Aratz: If e.g. the wrong version is specified during a release, it will be added to this file and it will need to be edited manually. There is also a risk it diverges from what's in `docs/` if these folders are edited or deleted manually. I would suggest to always regenerate the whole `switcher.json` from the versions available instead, e.g. list the folders in `docs/`, only keep the names that look like versions, and put that into a json file.

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,116 @@
...
+        env:
+          VERSION: ${{ steps.ver.outputs.version }}
+          DOCS_BASE_URL: https://PixelgenTechnologies.github.io/pixelator
+        run: |
+          python - <<'PY'
```
Aratz: Using such a long python script directly in a github workflow makes it easy to break it and hard to test it. I would recommend any of these two alternatives instead:
  - Use standard unix tools, i.e. collect versions with `ls` + `grep`, assemble the json with `jq`. This becomes a compact bash one liner that is much less dependent on formatting issues, and relies on battle-tested tools.
  - or, if you'd prefer to stick to python, I'd put that into a script in for example in `docs/_scripts`

Aratz (commented):
Great work! I have a few comments I think you should look into before we merge this 😉

.github/workflows/deploy-docs.yml
karlmoresco: Good point. Thing is that some warnings are caused by codebase ambiguities such as different classes sharing the same name, and re-exported members. Some refactoring is required to resolve, which I believe is outside the scope of this project. I can create a ticket for this but in the mean time, what do you think is the best approach for merging this PR?

Aratz: Ah, good point too, I'd suggest fixing as much you can, and suppressing warnings that are out of scope (see https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-suppress_warnings). Maybe you can also create a ticket with the maintenance tag in linear so that we can address these in the future.

karlmoresco: All right. Sounds good. Thanks!

PR #457, opened 2026-09-04
DO NOT MERGE - random test

elhb:
## Description

# DO NOT MERGE THIS!

I'm testing to drop high read support umis that would otherwise lead to uint16 overflow error and loud error
These will not be present in a normal dataset.

---

> [!NOTE]
> **Medium Risk**
> Changes collapse output by silently excluding high-support molecules and alter core read-weight math; downstream counts may differ from input on edge-case datasets even though typical runs are unchanged.
> 
> **Overview**
> Independent collapse now **drops molecule rows whose `read_count` is ≥ 2¹⁶** before writing parquet (output schema still uses `uint16`), via new `_build_output_table` that joins corrected UMIs first so filtering cannot misalign rows. **Logged output stats** (`output_reads`, `output_molecules`) reflect what is actually streamed, not input totals.
> 
> **Internal UMI weight aggregation** switches shared-memory read weights from `uint16` to `uint64` and uses `np.add.at` so many molecules sharing one UMI cannot wrap before directional collapse.
> 
> **Reports** replace computed `output_reads` aliases with explicit fields plus legacy validators that default missing values to input counts; combine/aggregate paths include the new output fields. Tests cover overflow drops, alignment, and uint64 weights.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 866af36ce783079eae9c8950c88aed3e92aeeabc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/pixelator/pna/collapse/independent/collapser.py
+++ b/src/pixelator/pna/collapse/independent/collapser.py
@@ -789,6 +789,15 @@ class RegionCollapser:
                 uei=uei_data,
             ).drop(["molecule"])
 
+            if drop_count := new_data.filter(pl.col("read_count").ge(2**16)).height:
+                logger.info(
+                    "Dropping of %s entries with too many reads (>=2^16) for %s, markers %s",
+                    drop_count,
+                    self.region_id,
+                    marker_name,
+                )
+                new_data = new_data.filter(pl.col("read_count").lt(2**16))
+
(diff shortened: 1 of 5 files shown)


elhb:
just used for test run, closing

